PR #84, opened 2026-04-22
Make Migration Flow Atomic

### LINKED ISSUE #81: Tighten  the Security Of PCR0 attestation chain audit trail

### LINKED ISSUE #83: Improve The Migration Process By making it  Atomic

aruokhai:
- Improve Migration Flow by Making It Atomic, closes #83 
- Tighten the security of the audit trail by making it difficult for a malicious party to alter, closes #81 

Future Exploration:
 The rollback of a previous enclave is a very delicate and not robust process; it would be better not to stop the previous enclave, but rather utilise a separate enclave for validation of the correctness of the new enclave before stopping the old enclave and restarting the new enclave. 
 This would require parking additional CPU and memory for migration validation, which would sit idle during normal enclave operations and only get activated during migration. I believe this is a good tradeoff for robustness. However, costs has to be taken into account.


aruokhai:
## Minor operational notes (non-blocking)

### 1. Rollback with both `Rename` + `copyFile` failing

**File**: [mgmt/migrate.go:429-435](mgmt/migrate.go#L429-L435)

If `os.Rename(eifBackup, eifDest)` AND the `copyFile` fallback BOTH fail during rollback, we emit `CRITICAL` and return — but v3's process was already killed at line 423. Both enclaves are now dead with no EIF at `eifDest`.

Low probability since `eifBackup` and `eifDest` are same-directory (rename shouldn't fail cross-FS), but worth a runbook note. Alternatively attempt a best-effort restart with whatever EIF is present so the failure surfaces via the enclave rather than silent dead-host state.

---

### 2. Export-key retry on same boot fails PCR31 lock

**File**: [sdk/migrate.go:182](sdk/migrate.go#L182)

`handleExportKey` is one-shot per old-enclave boot because PCR31 gets locked at line 186. If mgmt crashes after a successful `callExportKey` but before the EIF swap, the retry path fails here with "PCR locked" — the `existingMigKey` resume branch in `handleMigrate` can't recover without a reboot of the old enclave (resets PCR31).

Document in migration runbook; optionally detect this error in step 5 and surface a clearer message ("old enclave has stale PCR31, restart it before retrying").

---

### 3. Stray `.new` file on `os.Rename` failure

**File**: [mgmt/migrate.go:481-485](mgmt/migrate.go#L481-L485)

If `validate` passes but `os.Rename(stagePath, binPath)` fails, the staging `.new` file is left on disk. Cleaned on next successful update, but consider a `defer` guarded by a success flag to remove it unconditionally.

---

### 4. No auth on `/migrate` in test mode

**File**: [mgmt/main.go](mgmt/main.go) route registration

`/migrate` has no auth — it's localhost-bound on `127.0.0.1:8443` and in production is gated by SSM Run Command (IAM). Fine for tests, but worth a one-line comment on the route registration noting "production gating lives in IAM/SSM, not here" so a future reader doesn't accidentally expose this behind a public load balancer.

---

### 5. Preflight validate scope

**File**: [mgmt/validate.go](mgmt/validate.go)

Preflight covers STS + SSM connectivity but not port-bind capability or AWS SDK version compatibility. The trust model in `atomicMgmtUpdate`'s comment says this is acceptable, but production deployments may want an external health check from e.g. SSM Run Command verifying `/mgmt/health` after the swap.

PR #132, opened 2026-06-08
Verify Predecessor attestation

aruokhai:
closes #130 

Other Changes (Minor):
- Renamed the enclave.yaml property `prefix` to `Deployment`
- Prevent  the enclave properties **App** and **Deployment** from being overridden


ghost (changes requested):
## Arkana Security Review — PR #132 "Verify Predecessor Attestation"

**Verdict: Request changes.** The core fix is correct and necessary — `verifyPCR31Commitment` was trusting unsigned CBOR, which is a real vulnerability. But the implementation has gaps that must be closed before merge.

---

### 🔴 CRITICAL — Must fix

**1. `time.Now()` will reject valid attestations after cert expiry (`runtime/attestation.go:210`)**

The issue spec (#130) explicitly says: *"Use the attestation timestamp for the `CurrentTime` parameter as we do not care about expiry."*

```go
res, err := nitrite.Verify(doc, nitrite.VerifyOptions{
    Roots:       attestationRoots,
    CurrentTime: time.Now(),  // ← WRONG
})
```

The attestation is generated at migration time and stored in SSM. The new enclave verifies it at boot — which could be hours or days later. AWS Nitro leaf certs have finite validity. If the cert expires between migration and next boot (maintenance window, outage, slow rollout), `time.Now()` makes `nitrite.Verify` reject a perfectly valid attestation, **bricking the enclave on a valid migration chain**.

Fix: parse the attestation payload first to extract the `Timestamp` field, then pass `time.Unix(0, int64(doc.Timestamp)*int64(time.Millisecond))` as `CurrentTime`. This is what the issue spec requires.

**2. Missing PCR0 cross-check against `MigrationPreviousPCR0` (`runtime/migrate.go:469-480`)**

Issue #130 says: *"Verify stored migration attestation document against `MigrationPreviousPCR0` before checking `PCR31`."*

The PR verifies the COSE signature (good) and checks PCR31 (good), but never verifies that `result.Document.PCRs[0]` matches the stored `MigrationPreviousPCR0`. The `VerifyPredecessorCommitment` function reads `prevPCR0` at line 451 but never passes it through for cross-validation.

This is a defense-in-depth requirement from the spec. Without it, an attacker who can write to SSM could replace the attestation document with one from a *different* legitimate enclave (valid COSE, valid Nitro root, but different PCR0) that happened to commit PCR31 to the target value.

Fix: pass `prevPCR0` into `verifyPCR31Commitment` and add:
```go
attestedPCR0, ok := result.Document.PCRs[0]
if !ok || hex.EncodeToString(attestedPCR0) != prevPCR0 {
    return fmt.Errorf("attested PCR0 does not match stored MigrationPreviousPCR0")
}
```

**3. SSM env override can bypass COSE verification (`runtime/runtime.go:216` vs `runtime/runtime.go:249`)**

`applyEnvOverrides` runs at L216, **before** `VerifyPredecessorCommitment` at L249. The override function (`runtime/environment.go:133`) has no blocklist — it will happily `os.Setenv("ENCLAVE_DEPLOYMENT", "dev")` if `/<dep>/<app>/env/ENCLAVE_DEPLOYMENT` exists in SSM.

The comment at `runtime/environment.go:30-34` claims: *"The deployment is baked into the EIF at build time (and measured into PCR0), so this security-critical check cannot be flipped at deploy time."* **This claim is false.** The SSM env overlay runs before verification and can override `ENCLAVE_DEPLOYMENT`.

Fix (pick one):
- Add a blocklist in `applyEnvOverrides` that skips `ENCLAVE_DEPLOYMENT` (and ideally all security-critical vars)
- Or: capture the deployment value *before* env overrides and use a separate non-overridable flag for `skipCOSEVerification`
- Or: move `VerifyPredecessorCommitment` before `applyEnvOverrides` (but check for side effects)

---

### 🟡 HIGH — Should fix before merge

**4. Breaking `prefix` → `deployment` rename with no backward compatibility (`cli/config.go:40`)**

The YAML tag changes from `yaml:"prefix"` to `yaml:"deployment"` with no fallback. An existing `enclave.yaml` with `prefix: prod` will:
1. Silently ignore `prefix: prod` (unknown field)
2. Default `Deployment` to `"dev"` (L168)
3. Build an EIF with `ENCLAVE_DEPLOYMENT=dev`
4. **Skip COSE verification in the resulting enclave** (via `skipCOSEVerification`)

This is a silent security degradation on upgrade. Fix: add a YAML fallback that reads `prefix` if `deployment` is empty, with a deprecation warning printed to stderr. Or: split this rename into a separate PR with a migration guide.

---

### 🟢 MEDIUM — Observations

**5. `attestationRoots` global is a prod-reachable test seam (`runtime/attestation.go:191`)**

`var attestationRoots *x509.CertPool` is exported (package-level) and used in prod. Consider build-tag isolation (`//go:build !testing`) or at minimum making it unexported and only settable via an `init()` in a `_test.go` file. Currently it IS unexported but any code in the `runtime` package can set it.

**6. `parseCOSEPayloadInsecure` returns `SignatureOK: false` (`runtime/attestation.go:228-239`)**

Correct, but both the secure and insecure paths return the same `*nitrite.Result` type. Future callers may forget to check `SignatureOK`. Consider returning a distinct type for the insecure path, or at minimum add a comment warning at the `verifyAttestationDoc` return site.

**7. Two orthogonal changes in one PR**

The `prefix` → `deployment` rename is entirely unrelated to attestation verification. These should be separate PRs for clean review, bisect, and rollback.

---

### ✅ What's good

- The core security fix is the right approach: verify COSE before trusting PCR values
- Excellent test coverage: signed docs, forged docs, untrusted CAs, wrong targets, dev bypass — all the right cases
- `commitPCR31` idempotency and lock semantics are solid
- The `parseCOSEPayloadInsecure` dev-mode bypass is well-documented and correctly isolated behind `skipCOSEVerification`

---

**⚠️ PROTOCOL-CRITICAL**: This PR touches migration attestation verification — the trust chain that ensures only authorized enclave code can receive secret material. Items 1-3 above MUST be fixed. Human sign-off required before merge.

🤖 Reviewed by Arkana (Claude Opus)

runtime/attestation_verify_test.go
```diff
@@ -0,0 +1,346 @@
...
+	t.Setenv("ENCLAVE_DEPLOYMENT", "prod")
+
+	myPCR0 := hex.EncodeToString(bytes.Repeat([]byte{0xcd}, 48))
+	prevPCR0Bytes := bytes.Repeat([]byte{0x99}, 48)
+	prevPCR0 := hex.EncodeToString(prevPCR0Bytes)
+	forged := buildForgedAttestation(t, predecessorPCRs(t, prevPCR0Bytes, myPCR0))
+
+	if err := verifyPCR31Commitment(forged, myPCR0, prevPCR0); err == nil {
+		t.Fatal("forged (unsigned) attestation with correct PCR31 must be rejected")
+	}
```
chris-ricketts: The test just checks for any error. It should verify that it contains the string "verify predecessor attestation".
https://github.com/ArkLabsHQ/enclave/blob/ff659a621b686f79933c6efeba774630567bf1d6/runtime/migrate.go#L482-L485

runtime/attestation_verify_test.go
```diff
@@ -0,0 +1,346 @@
...
+	myPCR0 := hex.EncodeToString(bytes.Repeat([]byte{0xef}, 48))
+	prevPCR0Bytes := bytes.Repeat([]byte{0x99}, 48)
+	prevPCR0 := hex.EncodeToString(prevPCR0Bytes)
+	att := buildSignedAttestation(t, predecessorPCRs(t, prevPCR0Bytes, myPCR0))
+	other := buildSignedAttestation(t, predecessorPCRs(t, prevPCR0Bytes, myPCR0))
+	useAttestationRoots(t, other.roots) // trust a different CA
+
+	if err := verifyPCR31Commitment(att.docB64, myPCR0, prevPCR0); err == nil {
+		t.Fatal("attestation from an untrusted CA must be rejected")
+	}
```
chris-ricketts: The test just checks for any error. It should verify that it contains the string "verify predecessor attestation".
https://github.com/ArkLabsHQ/enclave/blob/ff659a621b686f79933c6efeba774630567bf1d6/runtime/migrate.go#L482-L485

runtime/attestation_verify_test.go
```diff
@@ -0,0 +1,346 @@
...
+	myPCR0 := hex.EncodeToString(bytes.Repeat([]byte{0x11}, 48))
+	otherPCR0 := hex.EncodeToString(bytes.Repeat([]byte{0x22}, 48))
+	prevPCR0Bytes := bytes.Repeat([]byte{0x99}, 48)
+	prevPCR0 := hex.EncodeToString(prevPCR0Bytes)
+	att := buildSignedAttestation(t, predecessorPCRs(t, prevPCR0Bytes, otherPCR0))
+	useAttestationRoots(t, att.roots)
+
+	if err := verifyPCR31Commitment(att.docB64, myPCR0, prevPCR0); err == nil {
+		t.Fatal("commitment to a different PCR0 must be rejected")
+	}
```
chris-ricketts: The test just checks for any error. It should verify that it contains the string "attested PCR0 does not match stored MigrationPreviousPCR0".
https://github.com/ArkLabsHQ/enclave/blob/ff659a621b686f79933c6efeba774630567bf1d6/runtime/migrate.go#L492-L494

runtime/attestation_verify_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+	myPCR0 := hex.EncodeToString(bytes.Repeat([]byte{0xab}, 48))
+	prevPCR0 := hex.EncodeToString(bytes.Repeat([]byte{0x99}, 48))
+	wrongPCR0Bytes := bytes.Repeat([]byte{0x77}, 48) // doc attests a different PCR0
+	att := buildSignedAttestation(t, predecessorPCRs(t, wrongPCR0Bytes, myPCR0))
+	useAttestationRoots(t, att.roots)
+
+	if err := verifyPCR31Commitment(att.docB64, myPCR0, prevPCR0); err == nil {
+		t.Fatal("attestation whose PCR0 differs from MigrationPreviousPCR0 must be rejected")
+	}
```
chris-ricketts: The test just checks for any error. It should verify that it contains the string "previous enclave committed to a different target PCR0".
https://github.com/ArkLabsHQ/enclave/blob/ff659a621b686f79933c6efeba774630567bf1d6/runtime/migrate.go#L504-L506

runtime/attestation.go
```diff
@@ -187,6 +189,71 @@ func extractPCRFromAttestation(attestationDoc []byte, index uint) ([]byte, error
 	return pcr, nil
 }
 
+// attestationRoots overrides the trust anchor for verifyAttestationDoc; nil uses
+// nitrite's embedded AWS Nitro root. Tests set a synthetic CA pool.
+var attestationRoots *x509.CertPool
```
chris-ricketts: Global state is harder to audit as it could be mutated from anywhere in the same package. 

Prefer threading roots from call sites, e.g.:

```go
func verifyPCR31Commitment(attestDocB64, myPCR0, prevPCR0 string) error {
	return verifyPCR31CommitmentWithRoots(attestDocB64, myPCR0, prevPCR0, nil)
}

func verifyPCR31CommitmentWithRoots(attestDocB64, myPCR0, prevPCR0 string, roots *x509.CertPool) error {
	attestDoc, err := base64.StdEncoding.DecodeString(attestDocB64)
	if err != nil {
		return fmt.Errorf("decode attestation base64: %w", err)
	}
	result, err := verifyAttestationDoc(attestDoc, roots)
	if err != nil {
		return fmt.Errorf("verify predecessor attestation: %w", err)
	}
	// existing PCR0/PCR31 checks...
}
```

Tests call `verifyPCR31CommitmentWithRoots` with fake root CA.

chris-ricketts (changes requested):

chris-ricketts (approved):
LGTM

PR #135, opened 2026-06-10
Pin the live TLS leaf cert

aruokhai:
Pin the live TLS leaf cert to the fingerprint signed into the enclave's NSM
attestation (tlsKeyHash in user_data), closing a MITM gap: the HTTP client,
`enclave verify`, and `enclave curl` previously skipped TLS verification, so a
host could terminate TLS with its own cert, relay the genuine enclave's
attestation, and read/tamper the channel while the client reported "verified".
gRPC already pinned; this brings the other three paths to parity.

Client:
- Split into an unpinned bootstrap client (only GET /enclave/attestation) and a
  pinned client for everything else. verify() fetches the attestation, activates
  the pin, then runs PCR + key-binding checks over the pinned client, so a cert
  mismatch fails the handshake before any app request is sent.
- Extract a shared verifyLeafCertPin used by both the HTTP and gRPC transports;
  reject empty/all-zero hashes (fail closed).
- Add Options.StrictTLS (public CA/hostname validation on top of the pin) and
  Options.SkipKeyBinding (PCR0 + pin only); export PinnedHTTPClient.

CLI:
- `enclave curl`: attestation-verified by default (--expected-pcr0), routing the
  user's request over the pinned connection. --insecure opts out (warns on
  stderr); --strict-tls adds PKI validation; the two are mutually exclusive.
- `enclave verify`: pins the live cert to the attested fingerprint after the
  unpinned bootstrap and runs remaining checks over it.

Runtime:
- setCertFingerprint fails closed when the chain has no non-CA leaf, instead of
  publishing an all-zero binding clients would reject.
- Drop the unused MockCertFp escape hatch.

Refactor:
- Rename appKeyHash → signingKeyHash (field + SetSigningKeyHash setter) to
  parallel its sibling tlsKeyHash.
- Move tests to their source-matching files: pin_test.go → client_test.go +
  verify_test.go, cert_fingerprint_test.go → runtime_test.go.


ghost (approved):
## Arkana Code Review — PR #135: Pin the live TLS leaf cert

**Verdict: APPROVE** — this is a high-quality security hardening PR that closes a real MITM gap and adds significant defense-in-depth. Thorough, well-tested, correct fail-closed semantics throughout.

---

### What this PR does (verified against the diff)

**TLS cert pinning (#129):** Splits the HTTP client into an unpinned bootstrap client (only `GET /enclave/attestation`) and a pinned client for everything else. The pin is activated after attestation verification, before any app request is sent. A shared `verifyLeafCertPin()` function covers HTTP, gRPC, and CLI paths — reject empty/all-zero hashes, reject missing peer certs, SHA-256 compare the DER leaf.

**State-origin receipts (#131):** NSM-signed receipts bind the SSM state (KMSKeyID + all secret ciphertexts + storage DEK) to the enclave's identity. Boot classification (genesis/resume/migration) fails closed on partial or ambiguous state. The canonical CBOR + domain-separated SHA-256 state_root is deterministic and sensitive to any ciphertext swap.

**KMS policy posture verification (#131):** `verifyKeyPolicyPosture` goes beyond `policyAdmitsPCR0` — it now rejects un-gated `kms:Decrypt` (no PCR0 condition), rejects `kms:PutKeyPolicy` to non-root principals (unlocked) or anyone (locked), and treats `kms:*` / `*` wildcards as granting both.

**SSM overlay hardening:** `nonOverridableEnv` prevents the SSM env overlay from setting `ENCLAVE_DEPLOYMENT` or `ENCLAVE_APP_NAME`, closing the attack where an SSM writer sets `ENCLAVE_DEPLOYMENT=dev` to flip `skipCOSEVerification()`.

**Migration ordering fix:** `storePCR0WithAttestation` now writes the attestation doc *before* the PCR0 commit point, and `VerifyPredecessorCommitment` fails closed on a recorded PCR0 with a missing attestation (was: `return nil`).

---

### Security analysis — no issues found

1. **Bootstrap/pin split is correct.** The unpinned client only fetches `/enclave/attestation` (public, no secrets). The pin is set in `verify()` via `c.setPinHash(tlsKeyHash)` before any call to `verifyKeyBinding` or `verifyPCR0Chain` runs on the pinned `c.httpClient`. Cache hits in `ensureVerified` skip re-verify but the `pinHash` persists from the first verification. Pre-pin requests fail closed (`verifyLeafCertPin` rejects empty hash).

2. **`setCertFingerprint` fail-closed** (`runtime/runtime.go:978`). Previously returned `nil` on a CA-only chain, leaving an all-zero `tlsKeyHash` that clients would accept. Now returns an error. Tested in `TestSetCertFingerprint`.

3. **`MockCertFp` removed.** The test-only escape hatch that let a config value override the real cert fingerprint is gone. Good.

4. **State-origin `classify()` is presence-only, verification is deferred.** This is the right design — classification reads SSM cheaply, then the caller runs the expensive NSM-verified check the mode selects before decrypting anything.

5. **Migration transition receipt ordering** (`runtime/migrate.go:261`). Written *before* the `KMSKeyID` flip (the atomic commit point), so the successor can verify the handoff on first boot. PCR31 already committed at this point.

6. **Rollback-onto-self** (`state_origin.go:~line 145`). When `predecessorPCR0Hex == ownPCR0Hex`, PCR31 check is skipped (it commits to the failed forward target, not us). PCR0 + state_root coverage is the proof. Well-tested in `TestVerifyMigrationTransitionReceipt_RollbackOntoSelf`.

7. **`verifyAttestationDoc` signature change** — removed the `roots` parameter, reads from the package-level `attestationRoots` var (nil = AWS Nitro root in production, test CA in tests). Fine for single-process runtime use. Tests use `t.Cleanup` to restore.

---

### Test coverage — excellent

- `verifyLeafCertPin`: match, mismatch, no cert, empty hash, all-zero hash
- `PinnedHTTPClient`: real TLS handshake (httptest.NewTLSServer) with matching/mismatching fingerprints
- `setCertFingerprint`: CA-only chain errors, leaf sets fingerprint
- `verifyKeyPolicyPosture`: 25+ table-driven cases covering un-gated decrypt, wildcards, locked/unlocked postures, mixed principals
- State origin: determinism, sensitivity, missing ciphertext errors
- Receipt verification: genuine, forged, wrong PCR0/purpose/state_root
- Migration transition receipt: genuine, wrong predecessor, wrong PCR31, rollback-onto-self
- Startup classification: genesis, resume, migration, 7 failure modes
- `nonOverridableEnv`: overlay can't flip `ENCLAVE_DEPLOYMENT`
- Attestation verification: signed valid, forged rejected, untrusted signer, wrong PCR0, expired cert valid at timestamp
- Integration test extended with `kms_key_locked` check

---

### Items to address (non-blocking)

**1. Downstream `prefix:` → `deployment:` breakage (3 repos)**
The config loader defaults `Deployment` to `"dev"` when the field is empty. Old `enclave.yaml` files with `prefix: prod` will silently ignore the value and use `"dev"` — a **silent misconfiguration**. Found in:
- `ee2e-kv/enclave/enclave.yaml`
- `ee2e-kv/infra/enclave.yaml`
- `introspector-review/enclave/enclave.yaml`

Consider adding a backward-compat shim in `loadConfigAt()`:
```go
if cfg.Deployment == "" && cfg.Prefix != "" {
    cfg.Deployment = cfg.Prefix
    slog.Warn("enclave.yaml: 'prefix' is deprecated, rename to 'deployment'")
}
```

**2. `InsecureTLS` default semantic change**
The old default for `client.Options.InsecureTLS` was effectively `true` (skip TLS verification). The new default is `nil` (pin to attested TLS hash). This is the correct direction, but any consumer calling `client.New()` without explicitly setting `InsecureTLS` will get different behavior. No external Go consumers were found in the monorepo, but this should be called out in release notes.

**3. `attestationRoots` global var thread safety** (`runtime/attestation.go`)
The package-level `var attestationRoots *x509.CertPool` is written by tests via `useAttestationRoots()` and read by `verifyAttestationDoc()`. Tests use `t.Cleanup` to restore, but parallel test execution could race. Consider adding `t.Parallel()` guards or documenting that attestation tests must not run in parallel. Non-blocking since the runtime is single-process and tests currently pass.

---

### Cross-repo impact

- **No external Go consumers** of `client.New()`, `client.Options`, `SetAttestationKeyHash`, or `MockCertFp` found across all repos in `/root/arkana/repos/`.
- **PR #132** (verify_attestation) and **PR #133** (enclave_restart) overlap on `runtime/runtime.go`, `runtime/attestation.go`, and `cli/curl.go` — expect merge conflicts.
- No open PRs on downstream consumer repos.

---

### Not protocol-critical

This PR does not touch VTXOs, transaction signing, forfeit paths, round lifecycle, or exit paths. It is enclave framework security — important, but not protocol-level. No protocol-critical human review flag needed.

🤖 Reviewed by Arkana

cli/verify.go
chris-ricketts, line 1: Rather than duplicate the sensitive attestation verification and TLS pinning logic in both this file and the `client` package, we should remove the helpers in this file and delegate to the `client` package.

e.g.:

```go
c, err := client.New(baseURL, client.Options{
	ExpectedPCR0:   expectedPCR0,
	StrictTLS:      strictTLS,
	SkipKeyBinding: !doVerifyKey,
})
if err != nil {
	return err
}

// runAllChecks is reduced to:
runCheck := verifyPCR0Chain(c)
```

 The `curl` command already does this.

client/client.go
```diff
@@ -52,10 +53,18 @@ type Options struct {
 	// Set to 0 to verify on every request. Default: 60s.
 	CacheTTL time.Duration
 
-	// InsecureTLS skips TLS certificate verification. Default: true.
-	// Nitriding uses self-signed certificates, so this is typically needed.
+	// InsecureTLS, when explicitly true, disables TLS cert pinning (raw,
+	// unverified TLS). Default pins to the attested tlsKeyHash (#129).
```
chris-ricketts: Downstream consumers of the `client` package don't really need to know the issue number (`#129`) or see it in the doc comments.

Similar with ``Set by `enclave curl` (#129)`` for `SkipKeyBinding`

cli/verify.go
chris-ricketts: Nit: the actual `waitSec` is always a multiple of the hardcoded interval `5s` 

e.g.: `waitSec = 6` and it doesn't connet/verify, the error will be: `verification failed after 6s: ...`

but it really waits for `10s`

runtime/runtime.go
```diff
@@ -938,57 +938,90 @@ func (e *Runtime) configureACME() error {
...
-			cancel()
-			if err == nil {
-				raw = got
-				break
-			}
-			time.Sleep(5 * time.Second)
+	// autocert rewrites the cache on every renewal, so the attested fingerprint
+	// must track the cache rather than sample it once.
+	go e.watchACMECert(cache)
+	return nil
```
chris-ricketts: Instead of watching the cache, we could wrap `mgr.GetCertificates` which writes to the cache:

```go
e.tlsGetCert = func(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
	cert, err := mgr.GetCertificate(hello)
	if err == nil && cert != nil && len(cert.Certificate) > 0 {
		// update hash if leaf cert changed
	}
	return cert, err
}
```

https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.48.0:acme/autocert/autocert.go;l=250

PR #138, opened 2026-06-26
feat(runtime): correct enclave clock drift via hypervisor PTP clock

### LINKED ISSUE #137: Fix enclave clock drift: sync CLOCK_REALTIME to the hypervisor PTP clock (/dev/ptp0)
## Problem

An AWS Nitro Enclave's clock (`kvm-clock`) is set once at boot from the Nitro Hypervisor and **never re-synchronized afterward**; it drifts ~1s/day proportional to load (AWS-confirmed). There is no time synchronization in the codebase, so the drift silently corrupts every wall-clock decision the runtime makes:

- **Freshness anchor** rollback records — `time.Now().Unix()` + S3 Object-Lock `RetainUntilDate` (`runtime/anchor.go`)
- **PCR0 lineage** audit timestamps + Object-Lock retain-until (`runtime/lineage.go`)
- **Self-signed TLS cert** `NotBefore`/`NotAfter` (`runtime/runtime.go`)
- **K/V TTL** expiry, **stream XADD** IDs, **migration cooldown** (`runtime/kvstore.go`, `runtime/resp.go`, `runtime/migrate.go`)

## Fix

Sync `CLOCK_REALTIME` to the hypervisor's PTP hardware clock (`/dev/ptp0`, the built-in `ptp_kvm` driver) for the enclave's lifetime:

- `(*Runtime).StartClockSync()` (`runtime/runtime.go`) opens `/dev/ptp0`, **hard-steps** `CLOCK_REALTIME` to it once at boot, then spawns the slew loop. Wired into `runtime/cmd/runtime/main.go` **before** any outbound TLS / listeners.
- `runClockSync()` (`runtime/clocksync.go`) reads the PTP + system clocks every 5s and, when drift exceeds 5ms, **slews gradually** via `unix.ClockAdjtime` (`ADJ_OFFSET_SINGLESHOT`) — never a backward jump, safe for the runtime's near-monotonic timestamp consumers. Exits on the lifecycle `stop` channel.
- Uses already-vendored `golang.org/x/sys/unix` (`ClockGettime`/`ClockSettime`/`ClockAdjtime`) — **no new dependency, no vendorHash change**. Runs as root → `CAP_SYS_TIME` present.
- **Graceful no-op** if `/dev/ptp0` is absent (logs `clock sync disabled: /dev/ptp0 unavailable` and returns).

## Why this is host-safe

The enclave's time comes from the Nitro **Hypervisor** (`/dev/ptp0`/`kvm-clock`), which sits outside the parent EC2 instance's trust boundary. A malicious host **cannot forge or rewind** the clock — setting the parent OS clock has zero effect, and it can't substitute a fake PTP device (the enclave kernel reads it via a KVM hypercall the host isn't in the path of). Across restarts the new enclave re-reads the hypervisor's current time. The host's only residual is **delay/DoS**, already in the threat model.

## Feasibility (verified)

The EIF's kernel is `aws/aws-nitro-enclaves-cli` **v1.2.3** (pinned via `monzo/aws-nitro-util`). Its `bzImage.config` has `CONFIG_PTP_1588_CLOCK_KVM=y` (built-in), `CONFIG_PTP_1588_CLOCK=y`, `CONFIG_KVM_GUEST=y`, and `CONFIG_DEVTMPFS_MOUNT=y` — so `/dev/ptp0` is auto-created at boot on real Nitro. No rootfs/kernel/mknod work needed.

aruokhai:
Fixes #137.

## Problem

The Nitro enclave's `kvm-clock` is set once at boot from the hypervisor and **never re-synchronized**, drifting ~1s/day. With no time sync in the codebase, that drift silently corrupts every wall-clock decision the runtime makes — freshness-anchor timestamps + S3 Object-Lock retain-until, PCR0 lineage, self-signed TLS validity, K/V TTLs, stream IDs, migration cooldown.

## Fix

Sync `CLOCK_REALTIME` to the hypervisor's PTP hardware clock (`/dev/ptp0`, the built-in `ptp_kvm` driver) — which the parent EC2 instance **cannot forge or skew**, even across restarts.

- `(*Runtime).StartClockSync()` opens `/dev/ptp0`, **hard-steps** `CLOCK_REALTIME` to it at boot (wired in `main.go` before any outbound TLS / served traffic), then spawns the slew loop.
- `runClockSync()` slews every 5s when drift exceeds 5ms via `unix.ClockAdjtime` (`ADJ_OFFSET_SINGLESHOT` — gradual, never a backward jump), and releases the device on the lifecycle `stop` channel.
- Uses already-vendored `golang.org/x/sys/unix` — **no new dependency, no vendorHash change**. **Fatal failure** if `/dev/ptp0` is absent.

## Why it's host-safe

Time is sourced from the Nitro **Hypervisor** (`/dev/ptp0`), outside the parent instance's trust boundary. A malicious host can't forge or rewind it — setting the parent OS clock has no effect, and it can't substitute a fake PTP device. Residual is delay/DoS only, already in the threat model.

## Feasibility (verified)

The pinned kernel (`aws-nitro-enclaves-cli` v1.2.3) has `CONFIG_PTP_1588_CLOCK_KVM=y` + `CONFIG_DEVTMPFS_MOUNT=y`, so `/dev/ptp0` is auto-created at boot on real Nitro — no rootfs/kernel work.

## Verification

- Full QEMU + LocalStack e2e: **35/35 integration, 76 run.sh PASS, 0 fail** — no regression.
- The KVM-accelerated runner exposes `/dev/ptp0`, so the boot hard-step actually engaged (`clock sync: initial hard-step to hypervisor PTP completed`); `run.sh` surfaces this as an informational line. A TCG-only runner would hit the graceful "disabled" branch.
- Unit tests for the offset math (overflow-safe across the 1e9 boundary) and the PTP clock-id derivation.

## Notes

- Base is `add_migration_attestion` (stacked on #136), since the clock-sync runs inside the same runtime boot path.
- Out of scope (separate concern): *proving* the time to a third party via hypervisor-signed, nonce-bound NSM timestamps (`tx.offchainTime`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


runtime/clocksync_linux.go
```diff
@@ -0,0 +1,275 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			slog.Info("clock sync stopping")
+			return
+		case <-ticker.C:
+		}
+		if err := cs.step(); err != nil {
+			slog.Error("clock sync: step failed", "error", err)
+		}
```
chris-ricketts, line 190: I think the clock sync errors should be hard errors instead of logged. They should be surfaced to the runtime so it does not continue with an un-synced clock.

e.g.

```go
// in StartClockSyncer
runCtx, cancel := context.WithCancelCause(ctx)

go func() {
  if err := cs.run(runCtx); err != nil {
     cancel(err)
  }
}()

return runCtx, nil

// make the run signature: 
func (cs *clockSyncer) run(ctx context.Context) error {
   // run loop, return any errors
}

// in runtime.Run start the sync and use the returned context
ctx, err := StartClockSyncer(ctx)
if err != nil {
  // ...
}
```

runtime/devcmdline.go
chris-ricketts, line 1: I think we can avoid the need for this entirely by using a test app endpoint:

1. Add `POST /test/clock-skew` to the test app; it advances `CLOCK_REALTIME` by 500 ms and returns the guest timestamp.
2. Call it from the existing integration test and confirm the guest is approximately 500 ms ahead of the host.
3. Wait slightly over the existing 5-second dev polling interval.
4. Query a test-app time endpoint and confirm guest time is again within a generous host-time bracket.
5. Assert the runtime logged one PHC recovery hard-step.

The following can then be removed:
- `runtime/devcmdline.go`
- `runtime/devcmdline_test.go`
- `ENCLAVE_CLOCK_TEST_STEP_NS`
- `maybeInjectCustomTime`
- `QEMU_APPEND` plumbing
- `ApplyDevCmdlineOverrides()` in main.go

chris-ricketts (changes requested):

runtime/devcmdline.go
aruokhai: Yep, that seems way easier.

runtime/clocksync_linux.go
```diff
@@ -0,0 +1,304 @@
...
+	for {
+		err := stepOnce()
+		if err == nil {
+			return nil
+		}
+
+		// PHC read failures are usually transient rather than indicating a
+		// dead clock. During ENA resets the AWS driver returns EOPNOTSUPP
+		// while the PTP device is reinitializing (/dev/ptp0 remains valid),
+		// and EBUSY during the subsequent ~1 ms back-off.
```
chris-ricketts, line 220: `/dev/ptp0` inside the enclave is the KVM virtual PTP clock, not an ENA PHC.

```suggestion
		// The enclave's /dev/ptp0 is the KVM virtual PTP clock. Its driver
		// reports EOPNOTSUPP when KVM_HC_CLOCK_PAIRING fails without
		// invalidating the device. Treat synchronization failures as potentially
		// transient and retry until the deadline.
```

- [KVM PTP maps clock-pairing failures to `EOPNOTSUPP`](https://github.com/torvalds/linux/blob/v4.14/drivers/ptp/ptp_kvm.c#L129-L136)

runtime/clocksync_linux.go
```diff
@@ -0,0 +1,304 @@
...
+// time synchronization service, so CLOCK_REALTIME would otherwise accumulate
+// drift over time.
+//
+// Sign convention: offset = PHC - CLOCK_REALTIME. A positive offset means
+// CLOCK_REALTIME is behind the PHC, so a positive frequency correction speeds
+// it up.
+//
+// References:
+//   - PI controller: https://en.wikipedia.org/wiki/PID_controller
+//   - linuxptp PI servo: https://github.com/richardcochran/linuxptp/blob/master/pi.c
```
chris-ricketts, line 43: Might be worth being more concise or readers will skip it.

```suggestion
// The clock syncer disciplines CLOCK_REALTIME against the trusted Nitro
// hypervisor's KVM virtual PTP clock) with a PI servo.
//
// Each cycle measures offset = PHC - CLOCK_REALTIME. The proportional term
// reacts to the current offset; the integral term retains the standing
// frequency correction. Positive offset means speed up CLOCK_REALTIME; large
// offsets are stepped.
//
// PI servo reference: https://github.com/richardcochran/linuxptp/blob/9db3cf9473eb5ba4f58fa269727c44793ae420d7/pi.c#L125-L150
```

runtime/clocksync_linux.go
```diff
@@ -0,0 +1,304 @@
...
+	retryInterval  time.Duration
+	failureTimeout time.Duration
+
+	cfg                    servoConfig
+	integralPpm            float64 // integral term: the standing frequency correction
+	lastXMonoNs            int64
+	hasPreviousMeasurement bool
+}
+
+func StartClockSyncer(ctx context.Context) (context.Context, error) {
```
chris-ricketts, line 111: I think it might be worth adding conditional hardening that verifies the current-clock source as suggested [here](https://blog.trailofbits.com/2024/09/24/notes-on-aws-nitro-enclaves-attack-surface/#time)

```go
// in environment.go
func verifyClockSourceEnabled() bool {
	return strings.EqualFold(strings.TrimSpace(os.Getenv("ENCLAVE_VERIFY_CLOCK_SOURCE")), "true")
}
```

```suggestion
func StartClockSyncer(ctx context.Context) (context.Context, error) {
	if verifyClockSourceEnabled() {
		data, err := os.ReadFile("/sys/devices/system/clocksource/clocksource0/current_clocksource")
		if err != nil {
			return nil, fmt.Errorf("read current clocksource: %w", err)
		}
		if clocksource := strings.TrimSpace(string(data)); clocksource != "kvm-clock" {
			return nil, fmt.Errorf("current clocksource is %q, want kvm-clock", clocksource)
		}
	}
```

This can be tested in E2E by checking that boot fails when the EIF is built with kernel params `no-kvmclock no-kvmclock-vsyscall` and baked-env `ENCLAVE_VERIFY_CLOCK_SOURCE=true`

runtime/clocksync_linux.go
```diff
@@ -0,0 +1,304 @@
...
+	cs.hasPreviousMeasurement = true
+
+	tx := unix.Timex{
+		Modes:  unix.ADJ_FREQUENCY | unix.ADJ_STATUS,
+		Status: unix.STA_FREQHOLD,
+		Freq:   ppmToKernelFreq(applied),
+	}
+	if _, err := clockAdjtime(unix.CLOCK_REALTIME, &tx); err != nil {
+		return err
+	}
```
chris-ricketts, line 277: State is updated before `clockAdjtime` succeeds. A 6 ms offset over 5 minutes gives `6,000,000 / (300 × 1,000) = 20 ppm`, so the PI output is `0.5 × 20 + 0.05 × 20 = 11 ppm`. If the syscall fails, another 10 seconds at 20 ppm adds 0.2 ms. Because the baseline was advanced, the retry calculates `6,200,000 / (10 × 1,000) = 620 ppm`, clamping the output to `100 ppm` and causing considerable overshoot.

We should commit the candidate integral and measurement baseline only after the adjustment succeeds.

```suggestion
	nextIntegralPpm := cs.integralPpm
	if correctingFreqPpm, ok := cs.frequencyErrorPpm(m); ok {
		nextIntegralPpm = clampFloat(cs.integralPpm+cs.cfg.ki*correctingFreqPpm, -cs.cfg.freqClampPpm, cs.cfg.freqClampPpm)
		proportionalPpm = cs.cfg.kp * correctingFreqPpm
		applied = clampFloat(proportionalPpm+nextIntegralPpm, -cs.cfg.freqClampPpm, cs.cfg.freqClampPpm)
	}

	tx := unix.Timex{
		Modes:  unix.ADJ_FREQUENCY,
		Freq:   ppmToKernelFreq(applied),
	}
	if _, err := clockAdjtime(unix.CLOCK_REALTIME, &tx); err != nil {
		return err
	}

	cs.integralPpm = nextIntegralPpm
	cs.lastXMonoNs = m.xMonoNs
	cs.hasPreviousMeasurement = true
```

runtime/clocksync_linux.go
```diff
@@ -0,0 +1,304 @@
...
+		applied = clampFloat(proportionalPpm+cs.integralPpm, -cs.cfg.freqClampPpm, cs.cfg.freqClampPpm)
+	}
+	cs.lastXMonoNs = m.xMonoNs
+	cs.hasPreviousMeasurement = true
+
+	tx := unix.Timex{
+		Modes:  unix.ADJ_FREQUENCY | unix.ADJ_STATUS,
+		Status: unix.STA_FREQHOLD,
+		Freq:   ppmToKernelFreq(applied),
+	}
```
chris-ricketts, line 274: `ADJ_STATUS` [replaces the kernel’s writable clock-status flags](https://github.com/torvalds/linux/blob/v4.14/kernel/time/ntp.c#L603-L613), so supplying only `STA_FREQHOLD` can clear unrelated flags such as `STA_INS`, `STA_DEL`, and `STA_UNSYNC`. This could cancel pending leap-second handling or temporarily mark an unsynchronized clock as synchronized.

Per [`adjtimex(2)`](https://man7.org/linux/man-pages/man2/adjtimex.2.html), `STA_FREQHOLD` only affects corrections made through `ADJ_OFFSET`, which we do not use. We should set only the frequency.

```suggestion
	tx := unix.Timex{
		Modes: unix.ADJ_FREQUENCY,
		Freq:  ppmToKernelFreq(applied),
	}
```

runtime/clocksync_linux.go
```diff
@@ -0,0 +1,304 @@
...
+
+	return &clockSyncer{
+		file:           file,
+		fd:             int(file.Fd()),
+		interval:       clockPollInterval(),
+		retryInterval:  clockSyncRetryInterval,
+		failureTimeout: clockSyncFailureTimeout,
+		cfg:            defaultServoConfig(),
+	}, nil
+}
```
chris-ricketts, line 171: Establish the monotonic baseline immediately after the initial hard-step. Currently, the first five-minute poll only initializes the baseline, delaying correction until ten minutes after startup. For a `20 ppm` drift, that first correction sees `12 ms / 5 min = 40 ppm`, doubling the actual drift estimate. With a startup baseline, the first poll correctly sees `6 ms / 5 min = 20 ppm`.

This also makes `hasPreviousMeasurement` redundant. It can be removed, with successful runtime hard-steps restarting the baseline by updating `lastXMonoNs`.

```suggestion
	var mono unix.Timespec
	if err := clockGettime(unix.CLOCK_MONOTONIC_RAW, &mono); err != nil {
		_ = file.Close()
		return nil, fmt.Errorf("establish initial clock baseline: %w", err)
	}

	return &clockSyncer{
		file:           file,
		fd:             int(file.Fd()),
		interval:       clockPollInterval(),
		retryInterval:  clockSyncRetryInterval,
		failureTimeout: clockSyncFailureTimeout,
		cfg:            defaultServoConfig(),
		lastXMonoNs:    unix.TimespecToNsec(mono),
	}, nil
}
```

runtime/clocksync_linux.go
```diff
@@ -0,0 +1,304 @@
...
+		// and EBUSY during the subsequent ~1 ms back-off.
+		if time.Now().After(deadline) {
+			return fmt.Errorf(
+				"clock sync failed for %s: %w",
+				cs.failureTimeout,
+				err,
+			)
+		}
+
+		slog.Warn("clock sync: read failed, retrying", "error", err)
```
chris-ricketts: Nit: it might not be a read error, could be from `adjust` or a write.

```suggestion
		slog.Warn("clock sync: synchronization attempt failed, retrying", "error", err)
```

runtime/environment.go
```diff
@@ -119,6 +127,20 @@ func getMigrationCooldown() time.Duration {
...
 
+func clockPollInterval() time.Duration {
+	if v := strings.TrimSpace(os.Getenv("ENCLAVE_CLOCK_POLL_INTERVAL")); v != "" {
+		if d, err := time.ParseDuration(v); err == nil && d > 0 {
+			return d
+		}
+	}
+	if IsDev() {
+		return 5 * time.Second
+	}
```
chris-ricketts: Dev builds can just set `ENCLAVE_CLOCK_POLL_INTERVAL`

I think we could remove `isDev()` completely and go back to just using `getDeployment() == "dev"` in `skipCOSEVerification()` unless we really want the extra `ENCLAVE_DEV` env var which is admittedly more explicit.

runtime/clocksync_linux_test.go
```diff
@@ -0,0 +1,377 @@
...
+		_ = cs.file.Close()
+	})
+}
+
+// TestStepRetry covers the retry/deadline logic that decides whether a transient
+// PHC read failure is tolerated or halts the runtime.
+func TestStepRetry(t *testing.T) {
+	origGet, origAdj := clockGettime, clockAdjtime
+	defer func() { clockGettime, clockAdjtime = origGet, origAdj }()
+	clockAdjtime = func(int32, *unix.Timex) (int, error) { return 0, nil }
```
chris-ricketts, line 326: We could have a test case that returns an error for `clockAdjtime` in order to test that the state is not updated, see: https://github.com/ArkLabsHQ/enclave/pull/138/changes#r3683664449

runtime/clocksync_linux_test.go
```diff
@@ -0,0 +1,377 @@
...
+	})
+
+	t.Run("ctx cancel during backoff is a clean stop", func(t *testing.T) {
+		clockGettime = func(int32, *unix.Timespec) error { return errors.New("EOPNOTSUPP") }
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		require.NoError(t, newCS().step(ctx))
+	})
+}
+
```
chris-ricketts, line 359: Add coverage for `measureOffset` to verify it calls the correct fds and the offset math, e.g.:

```suggestion
func TestMeasureOffset(t *testing.T) {
	original := clockGettime
	t.Cleanup(func() { clockGettime = original })

	const fd = 3
	phcID := fdToClockID(fd)

	var calls []int32
	clockGettime = func(clockID int32, ts *unix.Timespec) error {
		calls = append(calls, clockID)

		switch clockID {
		case unix.CLOCK_REALTIME:
			*ts = unix.Timespec{Sec: 100, Nsec: 200}
		case phcID:
			*ts = unix.Timespec{Sec: 100, Nsec: 500}
		case unix.CLOCK_MONOTONIC_RAW:
			*ts = unix.Timespec{Sec: 5, Nsec: 7}
		default:
			t.Fatalf("unexpected clock ID: %d", clockID)
		}
		return nil
	}

	got, err := (&clockSyncer{fd: fd}).measureOffset()
	require.NoError(t, err)
	require.Equal(t, offsetMeasurement{
		xMonoNs:  5*nsPerSecond + 7,
		phcNs:    100*nsPerSecond + 500,
		offsetNs: 300,
	}, got)
	require.Equal(t, []int32{
		unix.CLOCK_REALTIME,
		phcID,
		unix.CLOCK_MONOTONIC_RAW,
	}, calls)
}
```

chris-ricketts (changes requested):

runtime/environment.go
aruokhai: You are right, I prefer introducing an ENCLAVE_DEV environment variable. It makes it easier to configure different runtime constants for development versus production, especially for security- and correctness-critical operations. 

And by and by,  we should not have ENCLAVE_CLOCK_POLL_INTERVAL env variable

runtime/clocksync_linux_test.go
aruokhai: 7017b77b90a5f4e24600f7b526c22ac559460d8a

runtime/clocksync_linux_test.go
aruokhai: 7017b77b90a5f4e24600f7b526c22ac559460d8a

runtime/environment.go
aruokhai: 7017b77b90a5f4e24600f7b526c22ac559460d8a

runtime/clocksync_linux.go
aruokhai: 7017b77b90a5f4e24600f7b526c22ac559460d8a

runtime/clocksync_linux.go
aruokhai: 7017b77b90a5f4e24600f7b526c22ac559460d8a

runtime/clocksync_linux.go
aruokhai: 7017b77b90a5f4e24600f7b526c22ac559460d8a

runtime/clocksync_linux.go
aruokhai: 7017b77b90a5f4e24600f7b526c22ac559460d8a

runtime/clocksync_linux.go
aruokhai: 7017b77b90a5f4e24600f7b526c22ac559460d8a

aruokhai:
> **Status: large master merge (PR #144 + stacked migration commits) since last pass. 0 prior blockers fixed; R2 #4 partially fixed; 8 items still open.**
> 
> The entire incremental delta since `7017b77` is a merge of `origin/master` into `fix-clock_drift` (`66df42c`). The clock-sync files (`clocksync_linux.go`, `clocksync_linux_test.go`, `clocksync_darwin.go`) are unchanged. The migration-related additions (`state.go`, `migration_intent.go`, `servers.go`, `supervisor/migrate.go`, etc.) arrived wholesale from master and were reviewed through PR #144 separately — I am not re-auditing them here. I am auditing only whether the merge preserved or degraded the clock-sync invariants and whether any prior findings moved.
> 
> ## Prior findings — status
> ### R1 #1 — BLOCKER — `getDeployment()` default `"dev"` → `"test"` — **Still open, merge made it worse**
> `environment.go:66` (PR head) still returns `"test"` when `ENCLAVE_DEPLOYMENT` is unset. The master branch (which this merge pulled in) already has `return "dev"` — meaning the merge commit explicitly resolved the conflict in favour of the PR branch's `"test"`. Any deployment without an explicit `ENCLAVE_DEPLOYMENT` setting will silently route KMS, anchor, KV, and SSM paths to `/test/…` instead of `/dev/…`, orphaning all existing state. The fact that master reverted to `"dev"` and this merge kept `"test"` makes this harder to overlook — the two trees now diverge on this one line.
> 
> The `skipCOSEVerification()` decoupling from `getDeployment()` is correct and fully done (`IsDev()` / `ENCLAVE_DEV`), so the only reason to keep the default rename is the namespace isolation argument, which needs an explicit migration story. Please either revert `environment.go:66` to `"dev"` or document the migration path and confirm no live deployment relies on the implicit default.
> 
> ### R1 #2 — Should fix — PR description contradicts code — **Still open**
> PR body still says "Graceful no-op if `/dev/ptp0` is absent" (reality: fatal at boot) and "`ADJ_OFFSET_SINGLESHOT` — gradual, never a backward jump" (reality: `ADJ_FREQUENCY` + unconditional `ClockSettime`). Unchanged.
> 
> ### R1 #3 (partial) — Should fix — Runtime hard-step can move `CLOCK_REALTIME` backward — **Still open (partial)**
> `clocksync_linux.go:adjust()` still applies an unconditional `ClockSettime` on any offset exceeding `maxStepNs`. No forward-only clamp or accelerated-slew-for-backward fallback was added. The Warn log remains as the sole mitigation. Unchanged from prior delta.
> 
> ### R1 #5 — Nice to have — Four package-level `var` test seams — **Still open**
> `openPTPDevice`, `clockGettime`, `clockAdjtime`, `clockSettime` are still package-level `var`s in `clocksync_linux.go`. No tests in `runtime` use `t.Parallel()` today; the data-race risk remains theoretical. Unchanged.
> 
> ### R1 #6 — Nice to have — PI gains fixed regardless of poll cadence — **Still open**
> `kp=0.5, ki=0.05` are still constants. At the 5-min production cadence the integral takes many hours to stabilise on true drift. Unchanged.
> 
> ### R2 #2 — Should fix — 60s failure budget may be short vs ENA reset tails — **Still open**
> `clockSyncRetryInterval=10s`, `clockSyncFailureTimeout=60s` → at most ~6 attempts before halt. `retryInterval`/`failureTimeout` are struct fields, so making them configurable remains trivial. Unchanged.
> 
> ### R2 #4 — Nice to have — `waitForRuntime` doesn't call `child.Stop()` on cause-canceled ctx — **Partially fixed**
> `supervise()`'s direct `ctx.Done` arm (`runtime.go:259-262`, PR head) now correctly calls `child.Stop()` and returns the clock-failure error before exiting. That is the improvement. However, the `waitForRuntime` path (`runtime.go:265-273`) — reached when `Halt()` or `ChildDone()` fires first — still returns `fmt.Errorf("runtime halted: %w", cause)` without calling `child.Stop()`. In the `Halt()→waitForRuntime→ctx.Done(clock-failure)` sequence the child misses its SIGTERM. Not a correctness blocker (OS cleans up) but inconsistent with the direct path.
> 
> ### R2 #5 — Nice to have — `/test/clock-skew` unconditional and unauthenticated — **Still open**
> `test/app/cmd/main.go:132` registers `POST /test/clock-skew` with no `ENCLAVE_DEV` guard. Still matches the pattern of other `/test/*` endpoints. Still cheap to fix with a one-line guard at the top of the handler.
> 
> ### R2 #6 — Trivial — per-retry `time.NewTimer` allocation — **Still open**
> `clocksync_linux.go:step()` still allocates a new `time.Timer` per retry. Trivial.
> 
> ## New findings on this delta
> ### 1. `ENCLAVE_DEV` and `ENCLAVE_VERIFY_CLOCK_SOURCE` missing from the nonOverridableEnv test — minor gap
> `environment.go:17-26` (PR head) correctly lists both keys in `nonOverridableEnv`. However, the `TestApplyEnvOverrides / skips non overridable keys` subtest (`environment_test.go`) does not assert on either key — it only exercises the keys that were present before this PR. A future refactor could accidentally drop them from the map without a test catching it. Consider adding both to the non-overridable subtest.
> 
> ### 2. Clock-sync ordering preserved through the merge — confirmed ✓
> `runtime.go:41-44` (PR head): `StartClockSyncer` is still first, before `StartNetorking`, servers, TLS, AWS, and everything else. The master merge did not disturb this ordering. ✓
> 
> ### 3. `supervise()` ctx.Done cause-cancellation improved by the merge — confirmed ✓
> The merged `runtime.go:259-262` correctly handles clock-failure context cancellation: `cause != context.Canceled → child.Stop() → return error`. This is sound — SIGTERM is sent before returning the error. ✓ (The `waitForRuntime` gap from R2 #4 remains, but is narrower.)
> 
> ### 4. Migration code from master — out of scope, not re-audited
> `state.go`, `migration_intent.go`, `servers.go`, `supervisor/migrate.go`, `cli/migration.go` and the accompanying tests are wholesale from PR #144 master merge. They have been reviewed through that PR's own process. I am not re-auditing them in this pass. If there is any integration concern where the migration paths depend on wall-clock time (freshness-anchor, Object-Lock retain-until, migration cooldown timers), those paths now run behind the clock-sync servo, which is the correct posture.
> 
> ## Request
> R1 #1 (`getDeployment()` default) remains the only item blocking merge from a clock-sync review perspective. Everything else is a recommendation. Please either revert the default to `"dev"` or ship an explicit migration runbook as part of this PR.

R1 #3 — won't fix.

Backward steps aren't the hazard here, and a forward-only clamp would leave the real one untouched while preventing convergence.

In-process timing is immune. Go's time.Now() carries a monotonic reading, so time.Since, t.Sub, timers and context deadlines ignore CLOCK_REALTIME entirely. Only comparisons against persisted timestamps are exposed — KV TTL, migration cooldown, Object Lock retain-until — and that's a short, enumerable list.

Direction analysis cuts the other way. offset = PHC − CLOCK_REALTIME, and the step assigns CLOCK_REALTIME = PHC — so a backward step only occurs when our clock was ahead, and it lands us on truth, not below it. While ahead, time.Now().Add(window) writes Object Lock retain-until dates too far out: over-protection, failing safe. Already-written dates are absolute and enforced by S3's clock, so a later local step cannot change them. Retention is only shortened by a clock running behind — corrected by a forward step. Same asymmetry for attestation freshness: ahead makes valid documents look stale (fail-closed), behind makes stale ones look fresh (fail-open). The dangerous direction is the one the finding doesn't restrict.

The fix trades a bounded error for an unbounded one. Steps only fire above 100 ms, so hitting one means a real anomaly. If our clock is 5 minutes ahead and we refuse to step, we slew at the 100 ppm clamp — roughly 35 days computing retain-until dates and freshness checks against a clock we know is wrong. This system needs accuracy, not monotonicity; its security controls depend on the wall clock being right.

Threat model doesn't motivate it. We step to the Nitro hypervisor PHC, not host-supplied time, so this isn't an attacker-controlled channel.

Note also that the initial hard-step to PHC at startup is already unconditional and may be arbitrarily large in either direction — no forward-only clamp survives it.

If we ever guard this, the right axis is step magnitude, not direction: a forward six-month jump — expiring Object Lock early, making stale attestations look fresh — is far more dangerous than any backward correction.

aruokhai:
**R1 #6 — won't fix.**

The premise is that fixed gains are mis-tuned across cadences. They aren't: the controller's input is already a rate, so the loop is cadence-normalised by construction. `frequencyErrorPpm` computes `offset / elapsed` in ppm, and the interval cancels exactly when applying `kp` or `ki`. Simulating 5 s and 5 min cadences gives **bit-identical trajectories in ppm-space**; only the nanosecond offsets scale with cadence. The gains are therefore dimensionally correct at any poll interval.

A longer cadence only increases the offset for a given rate error and wall-clock settling time. The loop converges in ~28 steps at either cadence: ~2.3 min in dev and ~140 min in production by simulation. 

More importantly, this bound does not depend on the tuning analysis. `adjust()` checks `|offset| > maxStepNs` every poll and hard-steps `CLOCK_REALTIME` to the PHC when exceeded, zeroing the offset while retaining the learned integral. Thus the offset cannot persist above **100 ms**; even accounting for one interval's accumulation, the worst case is ~115 ms at 50 ppm. Slow integral convergence therefore cannot cause unbounded error regardless of `kp`/`ki`.

Increasing `ki` does not improve wall-clock convergence without sacrificing stability. The only lever that actually changes wall-clock convergence is cadence. **5 min is deliberate**, matching Evervault's published `/dev/ptp0` practice cited at the constant.

aruokhai:
R2 #2 . The instance clock, /dev/ptp0 is ENA backed, Enclave needs ptp_kvm. Howver it is worth fixing for roubstness sake:
At 100 ppm pessimistic:
60 s → 6 ms
10 min → 60 ms (still inside maxStepNs)
1 hour → 360 ms
1 day → 8.6 s

Increase to 10mins  is tolerable and helps avoid unnecessary enclave failure that might be caused by flaky /dev/ptp0 device

aruokhai:
R2 #5 Not Fixing, It is a test app

chris-ricketts (approved):
LGTM

PR #140, opened 2026-07-06
Refactor: runtime package

chris-ricketts:
## Summary:

- Split `runtime.go` (1,303 → 363 lines) into `servers.go`, `tls.go`, `network.go`, `dek.go`, `ssm.go`, `ssm_ttl_cache.go`, `migrate.go`
- Renamed `viproxy_setup.go` → `network.go`
- Merged `imds.go` into `aws_clients.go`
- Merged `transport.go`, `middleware.go`, `runtime_handlers.go` into `servers.go`
- Replaced `acme.go` with `acme_cache.go`
- Merged `nitriding_config.go` into `config.go`
- Replaced `policy_builder.go` with `kms_policy.go`; verification of PRC0 gated actions now only allows previous and/or current PCR0
- Replaced `storage.go` with `dek.go`
- Replaced `signature.go`; PCR0 signature loading absorbed into environment/nsm
- Rewrote `nsm.go`; added attestation option plumbing
- Added interfaces between subsystems (`KVStore`, `RuntimeState`, `Migrator`, `DEK`, etc.) to hide struct fields across files
- Removed `go-chi/chi`; routing via `net/http.ServeMux`
- Added `nitriding/system_darwin.go` for local builds
- New tests: `nsm_test.go`, `kms_policy_test.go`, `logging_test.go`, `servers_test.go`, `migrate_test.go`, `dek_test.go`, `ssm_test.go`, `ssm_ttl_cache_test.go`, `kms_test.go`, `acme_cache_test.go`
- Consolidated RESP/kvstore test suites; shared fakes moved to `utils_test.go`
- Net: prod 10,562 → 9,945 lines; tests 5,960 → 6,544;

Incorporates changes from #133 and #135


runtime/servers.go
```diff
@@ -0,0 +1,639 @@
...
+	cfg Config,
+	nsm NSM,
+	metrics *Metrics,
+	logging *Logging,
+	tracing *Tracing,
+	signer AttestationSigner,
+	hashes AttestationHashes,
+	authToken string,
+) Servers {
+	loggingMW := metricsMiddleware(metrics)
```
aruokhai: shouldn't this be metricsMW ?

runtime/servers.go
```diff
@@ -0,0 +1,639 @@
...
+		profiler.HandleFunc("GET /debug/pprof/symbol", pprof.Symbol)
+		profiler.HandleFunc("POST /debug/pprof/symbol", pprof.Symbol)
+		profiler.HandleFunc("GET /debug/pprof/trace", pprof.Trace)
+		// Q: Should this be on the external HTTP server? Auth?
+		em.Handle("/enclave/debug/", http.StripPrefix("/enclave", profiler))
+	}
+
+	im := http.NewServeMux()
+	// Q: Do we need this?
+	im.HandleFunc("POST /enclave/hash", hashHandler(hashes))
```
aruokhai: redundant, left over from nitriding port.

runtime/servers.go
```diff
@@ -0,0 +1,639 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to load signature PCR0: %w", err)
+	}
+
+	sig, err := ssm.MayGet(ctx, sigPath)
+	if err != nil {
+		return fmt.Errorf("failed to load signature: %w", err)
+	}
+
+	// Q: What is this for? Do we still need it?
```
aruokhai: Maybe, During Deployment, there exists an option to sign the PCR0 of the EIF file being deployed. Maybe it should not be passed into the runtime, rather be a deploy-time artifact.

runtime/servers.go
```diff
@@ -0,0 +1,639 @@
...
+			s.rt.NotifyListenerError(fmt.Errorf("private listener: %w", err))
+		}
+	}()
+
+	go func() {
+		var lis net.Listener
+		var err error
+		if cfg.UseVsockForExtPort {
+			lis, err = vsock.Listen(uint32(cfg.ExtPort), nil)
+		} else {
```
aruokhai: Could you explain the purpose of this logic

runtime/servers.go
```diff
@@ -0,0 +1,639 @@
...
+
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(res)
+	})
+
+	return nil
+}
+
+func attestationMiddleware(signer AttestationSigner) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
```
aruokhai: My bad, wrong choice of Struct name and function, Maybe function name could be  `responseSignerMiddleware` and  parameter type `AttestationSigner` could be `AttestedSigner`

runtime/servers.go
```diff
@@ -0,0 +1,639 @@
...
+func configHandler(cfg Config) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		_, _ = fmt.Fprintln(w, cfg)
+	}
+}
+
+// attestationHandler serves NSM attestation binding TLS and response-signing key hashes.
+func attestationHandler(useProfiling bool, nsm NSM, hashes AttestationHashes) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// Q: Why does enabling profiling disable the attestation endpoint?
```
aruokhai: Enabling profiling exposes net/http/pprof, which can dump enclave memory (keys, secrets), so attestation is refused to avoid vouching for an enclave whose secrets are readable.

runtime/static_secret.go
```diff
@@ -2,268 +2,146 @@ package runtime
...
-		select {
-		case <-ctx.Done():
-			if lastErr != nil {
-				return fmt.Errorf("KMS secret loading timed out after %s (last error: %v)", timeout, lastErr)
+		// Legacy secrets may already be hex.
+		candidate := strings.TrimSpace(string(plaintext))
+		if len(candidate) == 64 {
+			if _, err := hex.DecodeString(candidate); err == nil {
+				addSecret(s, candidate)
+				continue
```
aruokhai: lets not worry about legacy secrets

runtime/tls.go
```diff
@@ -0,0 +1,281 @@
...
+func acmeClientForDirectory(directory, caPEM string) (*acme.Client, error) {
+	var dirURL string
+	switch {
+	case strings.HasPrefix(directory, "https://"):
+		dirURL = directory
+	case directory == "letsencrypt-staging":
+		dirURL = acmeStagingDirectoryURL
+	default:
+		return nil, nil
+	}
```
aruokhai, line 227: lets differentiate intentional disable from wrong directory,

```
switch {
case directory == "":
    return nil, nil   // genuinely disabled — the one case that should be silent
case strings.HasPrefix(directory, "https://"):
    dirURL = directory
case directory == "letsencrypt-staging":
    dirURL = acmeStagingDirectoryURL
default:
    return nil, fmt.Errorf("unrecognized ACME directory %q", directory)
} 
```

aruokhai (changes requested):

runtime/state_origin.go
```diff
@@ -34,143 +22,238 @@ const (
...
+
+	for _, s := range secrets {
+		param := secretCiphertextParam(s.Name, kms.KeyID())
+		if _, err := ssm.MustGet(ctx, param); err != nil {
+			return StartStateInvalid, fmt.Errorf(
+				"required static secret missing from SSM: %s - %w", param, err,
+			)
+		}
+	}
+
```
aruokhai: I do not get why these are necessary for StateOrigin?

aruokhai (changes requested):
We already have state Root, wondering why this is necessary

runtime/attestation.go
```diff
@@ -1,95 +1,63 @@
...
+	SetSigningKeyHash(h [sha256.Size]byte)
+	Serialize() []byte
 }
 
-// SetRegistrar wires the in-process nitriding enclave as the recipient
-// of the attestation key hash. Call this before Init.
-func (a *Attestation) SetRegistrar(r AttestationHashRegistrar) {
-	a.registrar = r
+// AttestationSigner owns the ephemeral secp256k1 response-signing key.
+type attestationSigner struct {
```
aruokhai: The Signer is used basically to sign an enclave response, I do not know if `attestationSigner` is a  good enough descriptive name

runtime/migrate.go
```diff
@@ -0,0 +1,295 @@
...
+)
+
+// migrationPCRIndex stores the successor-PCR0 handoff commitment.
+const migrationPCRIndex = 31
+
+type StartMigrationRequest struct {
+	NewPCR0 string `json:"new_pcr0"`
+}
+
+type StartMigrationResult struct {
```
aruokhai, line 21: I would love to get rid of this file, I think all its operations can naturally be methods of the Runtime struct

runtime/resp.go
```diff
@@ -202,160 +189,165 @@ func (s *RESPServer) handle(conn redcon.Conn, cmd redcon.Command) {
...
 		s.cmdXAdd(ctx, conn, cmd)
-	case "XLEN": // stream LENgth → entry count
+	case "XLEN":
 		s.cmdXLen(ctx, conn, cmd)
-	case "XRANGE": // stream RANGE of entries by ID [COUNT n]
+	case "XRANGE":
 		s.cmdXRange(ctx, conn, cmd)
-	case "XREAD": // stream READ entries after given IDs
+	case "XREAD":
 		s.cmdXRead(ctx, conn, cmd)
```
aruokhai: I actually prefer such explanatory comments to be here

runtime/servers.go
chris-ricketts: It is a port of this logic:

https://github.com/ArkLabsHQ/enclave/blob/4312260f12d0d642582c54ef25be36b15fac9b0b/runtime/runtime.go#L779-L784

Note: `UseVsockForExtPort` is never set and only ever has the default value of `false` currently.

What was the original intended use case? Can we remove it?

runtime/servers.go
chris-ricketts: Do we need to have `UseProfiling` at all? 

We are forcing the reader to verify that it is not enabled as it is dangerous, however the profiling is almost never used, i.e. only on local debug builds.

I suggest removing it entirely and patching the file for local debug builds if it is ever required.

runtime/static_secret.go
chris-ricketts: Port of logic from:

https://github.com/ArkLabsHQ/enclave/blob/4312260f12d0d642582c54ef25be36b15fac9b0b/runtime/static_secret.go#L253-L263

Agree we can assume that `GenerateDataKey` is always used to initialise the secrets.

runtime/state_origin.go
chris-ricketts: It is just an early failure if those values are missing as they are required to later validate any existing state root

runtime/migrate.go
chris-ricketts: I think it makes sense to have the sensitive migration logic grouped together in one file with an associated unit test file.

While there is no longer a Runtime struct - they could be free functions within the `runtime.go` file.

I think we should wait to see how it looks once we add the cooldown enforcement and one-shot anchor logic.

runtime/servers.go
chris-ricketts: e312677d65c967afb5ecf14b2ea2bd9e0f277f31

runtime/servers.go
chris-ricketts: 93e0fa45861e1e8983127179007cfa012692ea04

runtime/servers.go
chris-ricketts: 6821959e785f20470614fc4d7fa2b2670dfb45e2

runtime/static_secret.go
chris-ricketts: 48e3fa754aae266526fc630d7d9daa24233afa9d

runtime/tls.go
chris-ricketts: f14a1895fd731702da0296c4e70829ff4d224624

runtime/attestation.go
chris-ricketts: 6821959e785f20470614fc4d7fa2b2670dfb45e2

runtime/servers.go
aruokhai: yep, we can remove

runtime/state_origin.go
aruokhai: In so far it is validated somewhere else, and not a critical logic here, I reckoned that it can be removed here, so as to keep such function committed to one goal.

runtime/migrate.go
aruokhai: okay, okay

runtime/resp.go
aruokhai: can we have this back :)  ?

runtime/servers.go
aruokhai: Haha, my bad, we can remove it, It is a left over

runtime/resp.go
chris-ricketts: 03660f197b28f45b4085145d29096dce7a736e6c

runtime/state_origin.go
chris-ricketts: d2fc4f9300f584d665ab0e5334b691365fc8bd9d

runtime/servers.go
chris-ricketts: d5c16cdf7f7cc7d0708741ce3e2a7701afc06ef8

runtime/servers.go
chris-ricketts: 80f513960c14230d6fb42c8c5efd173696262114

runtime/servers.go
chris-ricketts: Removed from runtime and tofu SSM provision fefc0b6d6329988de8d64f6d767ff7a3331a206f

PR #144, opened 2026-07-16
Feat: enclave-enforced migration cooldown

chris-ricketts:
Closes #142


runtime/environment.go
```diff
@@ -107,16 +108,28 @@ func kmsKeyIDParam() string {
...
+}
+
+func migrationIntentRetention() time.Duration {
+	if value := strings.TrimSpace(os.Getenv("ENCLAVE_MIGRATION_INTENT_RETENTION")); value != "" {
+		if retention, err := time.ParseDuration(value); err == nil && retention > 0 {
+			return retention
+		}
+	}
+	return 10 * 365 * 24 * time.Hour
 }
```
aruokhai: Let us also make it a strict check like the getMigrationCooldown().

runtime/migrate.go
```diff
@@ -84,75 +111,155 @@ func (m *migrator) PreviousPCR0Info(ctx context.Context) (*PreviousPCR0Info, err
...
-		RemainingSeconds:  int(rem.Seconds()),
-		Pending:           true,
-	}, nil
+	eligibleAt := head.PublishedAt.Add(cooldown)
+	status.EligibleAt = &eligibleAt
+	remaining := eligibleAt.Sub(now)
+	if cooldown == 0 || remaining <= 0 {
+		status.State = migrationStateEligible
+		return status
+	}
```
aruokhai: cooldown == 0 check should happen before eligbleAt variable declaration, as it is not dependent on anything

runtime/migrate.go
```diff
@@ -84,75 +111,155 @@ func (m *migrator) PreviousPCR0Info(ctx context.Context) (*PreviousPCR0Info, err
...
 
-// StartMigration exports state under a PCR0-locked migration key, then flips KMSKeyID.
-func (m *migrator) StartMigration(
+// CompleteMigration exports state under a PCR0-locked migration key, then flips KMSKeyID.
+func (m *migrator) CompleteMigration(
 	ctx context.Context,
-	newPCR0 string,
-) (*StartMigrationResult, error) {
-	newPCR0Bytes, err := validateNewPCR0(newPCR0)
+	targetPCR0 string,
```
aruokhai: I reckon that this parameter is unnecessary now, we only have one targetPCR0 always active

runtime/migration_intent.go
```diff
@@ -0,0 +1,424 @@
...
+		}
+		if out.NextKeyMarker == nil && out.NextVersionIdMarker == nil {
+			return scan, fmt.Errorf("%w: list migration intents: truncated response missing markers", errMigrationIntentStoreUnavailable)
+		}
+		keyMarker, versionMarker = out.NextKeyMarker, out.NextVersionIdMarker
+	}
+}
+
+func (l *migrationIntentLog) readVersion(
+	ctx context.Context,
```
aruokhai: probably a better function name shoud be `getIntent`

runtime/migration_intent.go
```diff
@@ -0,0 +1,424 @@
...
+type migrationIntentHead struct {
+	SourcePCR0  string
+	TargetPCR0  string
+	Action      string
+	Sequence    uint64
+	PublishedAt time.Time
+	VersionID   string
+}
+
+type migrationIntentScan struct {
```
aruokhai: This seems unnecessary, I reckon what is needed is only the head, let us drop the struct

runtime/migration_intent_test.go
```diff
@@ -0,0 +1,566 @@
...
+		publishedAt := time.Now().Add(-time.Hour).UTC()
+		fx.s3.putRawObjectAt(key,
+			fx.object(t, 1, migrationIntentRequested, targetA, migrationIntentTestBucket, fx.pcr0),
+			publishedAt)
+		fx.s3.putRawObjectAt(key,
+			fx.object(t, 1, migrationIntentRequested, targetB, migrationIntentTestBucket, fx.pcr0),
+			publishedAt)
+
+		_, err := fx.log.Head(context.Background())
+		require.ErrorIs(t, err, errMigrationIntentAmbiguous)
```
aruokhai, line 256: Wrong type of check, when we get rid of the `scan` struct, we need to call the public function to insert a wrong request ( A new migration request when there is an existing migration request), and such call needs to fail, because the head is not an `AbortRequest`. We can mock the necessary types  (s3, nsm) methods, so we do not make strange calls.

runtime/migration_intent_test.go
```diff
@@ -0,0 +1,566 @@
...
+	}
+	return out, err
+}
+
+type readErrorCloser struct{ err error }
+
+func (r readErrorCloser) Read([]byte) (int, error) { return 0, r.err }
+func (readErrorCloser) Close() error               { return nil }
+
+func TestMigrationIntentS3Failures(t *testing.T) {
```
aruokhai, line 390: I dont think necessary, maybe if we mock struct methods, this can be removed

runtime/state.go
```diff
@@ -0,0 +1,610 @@
...
+
+	return verifiedState{
+		kms:                       kms,
+		dek:                       &dek{key: dekPlaintext},
+		secrets:                   secrets,
+		migrationIntentBucketName: snapshot.migrationIntentBucketName,
+	}, nil
+}
+
+func validateStaticSecretArtifacts(metadata []StaticSecretMetadata) error {
```
aruokhai: `verifyStatcSecretsExistence` , a more descriptive name

runtime/state.go
```diff
@@ -0,0 +1,610 @@
...
+	predecessorPCR0        string
+	predecessorAttestation string
+}
+
+type verifiedState struct {
+	kms                       PrimaryKMS
+	dek                       DEK
+	secrets                   []StaticSecret
+	migrationIntentBucketName string
+}
```
aruokhai, line 51: Trying to understand why this distinction, and not a single State struct

runtime/state.go
```diff
@@ -0,0 +1,610 @@
...
+	return verified, nil
+}
+
+func initializePersistedState(
+	ctx context.Context,
+	kms PrimaryKMS,
+	ssm SSM,
+	metadata []StaticSecretMetadata,
+	migrationIntentBucketName string,
+) (persistedStateSnapshot, verifiedState, error) {
```
aruokhai, line 377: I do not understand why need to return a separate persistedStateSnapshot struct

runtime/state_origin.go
```diff
@@ -1,361 +0,0 @@
...
-
-type StartState int
-
-const (
-	StartStateGenesis StartState = iota
-	StartStateResume
-	StartStateMigration
-	StartStateInvalid
-)
-
```
aruokhai, line 33: would have been to have refactor as a separate PR

test/run.sh
```diff
@@ -644,533 +680,818 @@ fi
...
+assert_no_legacy_receipt "$K_AC"
+assert_host_migration_cleanup
+echo "  PASS: A-as-C failed closed, retained K-AC/A, and did not promote the supervisor"
+
+PRE_REAL_C_SUP_PID=$(cat "$SUPERVISOR_PIDFILE")
+migration_finalise --resume \
+  --target-pcr0 "$V4_PCR0" \
+  --artifact-bucket "$V4_CANDIDATE_BUCKET" \
+  --eif-key "$V4_CANDIDATE_EIF_KEY" \
+  --supervisor-key "$V4_CANDIDATE_SUPERVISOR_KEY" \
```
aruokhai, line 1347: I believe we talked about not including keys, right ?

aruokhai (commented):
The core changes are correct

runtime/migrate.go
chris-ricketts: The `targetPCR0` parameter is intended to make the state change request explicit.

E.g. Current head is PCR0 `A`. The eligible migration is to PCR0 `B`, however the operator builds an EIF with PCR0 `C`, which is the desired migration target. If they specify PCR0 `C` in the `/finalize-migration` then it fails early and no state changes are made: new KMS key, re-encryption, migration receipt, etc.

Without the guard, the runtime implicitly migrates to `B` and there is more cleanup / work to rollback to PCR0 `A` and re-request a migration to PCR0 `C`.

runtime/migration_intent.go
chris-ricketts: The `maxSequence` field could be dropped as it's always equal to the `head.Sequence` or zero if `head` is `nil`. `ambiguous` is still required to handle the edge case of a race on S3 producing same-sequence log entries with the exact same timestamp — highly unlikely but technically possible.

IMO It'd be cheaper (LoC wise) and easier to read if we keep the struct than thread this state through a load of free functions.

runtime/migration_intent_test.go
chris-ricketts: See https://github.com/ArkLabsHQ/enclave/pull/144#discussion_r3664863470 - regarding keeping the struct.

The checks already use the public API: `Head` and `Request`. Only the setup writes to the fake S3 directly. That part is unavoidable as the test needs two entries with the same sequence and the exact same timestamp, and `Request` always appends at `maxSequence+1`, so public calls can never create that state.

A new request while one exists does not fail. It supersedes at a higher sequence. 
That test confirms: fail closed on an exact tie, then let a later request append past it.

runtime/migration_intent_test.go
chris-ricketts: Which part do you mean: `TestMigrationIntentInvalidVersionsAreIgnored` or the `migrationIntentS3Failure` decorator? They are separate concerns.

runtime/state.go
chris-ricketts: They're three phases of the same state at different trust levels. `unverifiedState` is what we load from SSM before the receipt verifies. `verifiedState` only exists after, holding the live KMS and plaintext. One struct would be half empty at every phase.

`persistedStateSnapshot` is separate because it defines exactly what the state-origin receipt commits to. Those four fields are the `state_root` hash input. It also gets built by three different paths: loaded at boot, generated at genesis, and constructed by the predecessor during handoff (`migrate.go:332`), where there is no unverified state at all.

Genesis (`initializePersistedState`) returns both because it just created both: ciphertext for the receipt hash, plaintext for use. Returning only the snapshot would mean KMS-decrypting values we already have.

I can add some comments to make those distinctions more obvious.

runtime/state.go
chris-ricketts: see https://github.com/ArkLabsHQ/enclave/pull/144#discussion_r3665220741

runtime/state.go
chris-ricketts: 2994e8c7a127bb28102c30406a4a78c5d1b4a548

runtime/state.go
chris-ricketts: I renamed to `validateStaticSecretNames` as it's validating there is not a collision with the reserved name `StorageDEK` or duplicate names, rather than checking for existence.

303356fa2a0134749e778b5f5541a0c3f2d6d78c

runtime/migration_intent.go
chris-ricketts: renamed to `fetchIntentHead` - in my mind, `get` implies infallible
1944e076bcc2349da3def3b08213662b361a7cd4

runtime/migrate.go
chris-ricketts: [a474cdc1c54862bb8fc05c37ef5a570919256a6e](https://github.com/ArkLabsHQ/enclave/pull/144/commits/a474cdc1c54862bb8fc05c37ef5a570919256a6e)

runtime/environment.go
chris-ricketts: [cdb11020121bcbd29480aad8efd50c102caab504](https://github.com/ArkLabsHQ/enclave/pull/144/commits/cdb11020121bcbd29480aad8efd50c102caab504)

test/run.sh
chris-ricketts: I think it is better to be explicit. This script juggles four candidate versions and half the calls target non-current artifacts (the A-as-C negative test passes retained A keys while tofu points at C). Uniform explicit flags keep every call site self-contained instead of depending on what tofu happens to point at.

runtime/migration_intent.go
chris-ricketts: f54d5334b2fa02513e67a9b912cb01f988fe191b

runtime/migrate.go
aruokhai: After a migration request to `B`, the head of the log points to a migration targeting `B`, so FinalizeMigration simply completes that migration. It is more sensible to build before migrating, abort if a different PCR0 is required, and then issue a new migration request. Restricting the system to a single migration target at any given time makes the state machine much easier to reason about and simplifies verification.

chris-ricketts: > Restricting the system to a single migration target at any given time makes the state machine much easier to reason about and simplifies verification.

This is still the case but it's forcing the caller of `/finalise-migration` to confirm their intent by matching the eligible PCR0 in the request rather than implicitly accepting whatever migration is eligible.

It's not a big deal though - I can remove the check

runtime/migration_intent.go
aruokhai: Is it crucial to persist the `ambiguous` variable? Is there any downside to failing fast instead?

chris-ricketts: This is how it looks without the `migrationIntentScan` struct: 886a757c81da8772a427a0c74f225dd9ac5edb44

It can't fail fast on `ambiguous` as there may be valid versions with higher sequences that override ambiguity in lower sequences. 

Hard to say which one is more readable - without `migrationIntentScan` is a bit more fiddly across functions but is less LoC. 

Let me know which you prefer.

runtime/migrate.go
chris-ricketts: 03fce8de03fdc39b79c4e02faabd38cd94d71e9f

runtime/migration_intent.go
chris-ricketts: https://github.com/ArkLabsHQ/enclave/blob/886a757c81da8772a427a0c74f225dd9ac5edb44/runtime/migration_intent.go#L260-L262

This is code smell as it returns a non-nil `head` and non-nil `err` that the caller needs to know how to handle.

I think this version is a winner: 0e5db57139f2b05fe2b7d7e65ee750e87841ef52

runtime/state.go
aruokhai: Yep, finally understood the flow

runtime/migration_intent_test.go
aruokhai: I was saying cant we mke s3 an interface, so we can create a mock implementation of methods to insert the data, which may read cleanly

chris-ricketts: Good shout: 247506aaff7cda5145be1231940f80a003f71641

PR #149, opened 2026-07-31
Refactor:  replace enclave framework with Nix runtime toolchain

chris-ricketts:
## Summary

Refocus this repository from a full application framework into a reusable Nix toolchain and Go runtime for AWS Nitro Enclaves. Downstream applications now package their own executable and consume the exported flake API directly.

## What changed

- Add public Nix functions for building measured EIFs, production AMIs, QEMU test hosts, and OpenTofu deployment stacks.
- Add a downstream application flake template.
- Move host lifecycle management into shared NixOS services used by both production Nitro hosts and QEMU test hosts.
- Generate a blue/green OpenTofu stack with a stable EIP, PCR0-specific AMIs, Object-Locked migration intents, SSM configuration, and least-privilege IAM.
- Retain the attestation-verifying Go client and replace the framework CLI with a focused `enclave curl` command.
- Rename Go modules from `github.com/ArkLabsHQ/introspector-enclave` to `github.com/ArkLabsHQ/enclave`.
- Fix ACME configuration through SSM, default-SNI handling, and encrypted cache key layout.
- Replace the Docker-based integration harness with NixOS tests that boot real measured EIFs under QEMU.
- Rewrite the documentation around the new architecture, deployment model, migration runbook, and security boundaries.

## Why remove the K/V store?

This removes roughly 5,600 lines, including tests, implementing the Redis-compatible K/V store, RESP server, and rollback anchoring.

We do not yet know whether downstream Arkade applications need this functionality. Keeping it in the core runtime would require every change to be maintained, reviewed, and security-audited despite having no confirmed consumer.

If downstream demand emerges, the storage engine can be extracted into a separate library. The runtime would export a DEK to the application, and the library would consume it alongside standard AWS configuration and storage resource names. Preserving rollback protection would additionally require an Object-Locked S3 anchor bucket.

## End-to-end coverage

The NixOS test exercises:

- EIF measurement and production AMI construction
- OpenTofu validation and blue/green plan invariants
- Genesis state and static-secret creation
- Attestation-verified application requests
- Blue-to-green migration and state adoption
- Stable EIP cutover and blue retirement
- Pebble ACME issuance and encrypted S3 certificate caching
- Green restart and certificate cache reuse

## Breaking changes

- The previous application scaffolding, deployment CLI, host supervisor, runner, and release workflows are removed.
- The built-in Redis-compatible, rollback-resistant K/V store is removed. It can be restored as a separate library if downstream demand materializes.
- The CLI now exposes only attestation-verified `enclave curl`.
- The Go module path changes to `github.com/ArkLabsHQ/enclave`.
- Applications must consume the Nix flake directly.
- BusyBox is no longer included in EIFs by default; applications requiring shell utilities must pass them through `extraPackages`.

## Validation

CI now runs:

- `nix develop --command make test`
- `nix develop --command make lint`
- `nix build .#checks.x86_64-linux.e2e --print-build-logs`

The full e2e check requires `x86_64` or `aarch64` Linux, nested KVM, and the `nixos-test` system feature.

Note: `aarch64` target has not been tested.

## Minimal downstream consumer `flake.nix` 

A downstream application only needs to package its executable and pass it to `buildEif`:

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    enclave.url = "github:ArkLabsHQ/enclave";
  };

  outputs =
    { nixpkgs, enclave, ... }:
    let
      system = "x86_64-linux";
      pkgs = import nixpkgs { inherit system; };

      app = pkgs.buildGoModule {
        pname = "myapp";
        version = "0.1.0";
        src = ./.;
        subPackages = [ "cmd/myapp" ];
        vendorHash = null;
        env.CGO_ENABLED = "0";
      };
    in
    {
      packages.${system}.default = enclave.lib.buildEif {
        inherit pkgs app;
        env = {
          ENCLAVE_DEPLOYMENT = "prod";
          ENCLAVE_APP_NAME = "myapp";
          ENCLAVE_AWS_REGION = "us-east-1";
          ENCLAVE_PREVIOUS_PCR0 = "genesis";
          ENCLAVE_MIGRATION_INTENT_RETENTION = "87600h";
        };
      };
    };
}
```

`buildEif` is the smallest integration point. Downstream flakes can also use `lib.mkEnclaveAmi` to build a Nitro-capable NixOS AMI and `lib.mkEnclaveTofu` to generate the blue/green AWS deployment definitions. `lib.mkEnclaveQemuAmi` provides the equivalent QEMU host for NixOS tests.


nix/tests/test-app/testapp
aruokhai: Think this needs to be deleted

nix/tests/e2e.nix
aruokhai: would love if maybe this file could be read like a real spec:

```json
{
  pkgs,
  self,
  aws-nitro-util,
  enclaveTofu,
  commonEifEnv,
  awsNodeIP,
  support,
}:

let
  lib = pkgs.lib;
  system = pkgs.stdenv.hostPlatform.system;

  inherit (support)
    testApp
    awsmocks
    ministack
    pebbleFixtures
    ;

  /*
   * Patched dependencies
   */

  eifInit =
    aws-nitro-util.packages.${system}.eif-init.overrideAttrs
      (oldAttrs: {
        patches =
          (oldAttrs.patches or [ ])
          ++ [ ./fixtures/patches/aws-nitro-util-init-wait.patch ];
      });

  vhostDeviceVsock =
    pkgs.vhost-device-vsock.overrideAttrs
      (oldAttrs: {
        patches =
          (oldAttrs.patches or [ ])
          ++ [ ./fixtures/patches/vhost-device-vsock-forward-listen.patch ];

        checkFlags =
          (oldAttrs.checkFlags or [ ])
          ++ [
            "--skip"
            "tests::test_vsock_server_vsock"
          ];
      });

  /*
   * Enclave images
   */

  mkEif =
    env:
    self.lib.buildEif {
      inherit pkgs;
      app = testApp;
      init = "${eifInit}/bin/init";

      env =
        commonEifEnv
        // { ENCLAVE_DEV = "true"; }
        // env;
    };

  blueEif = mkEif {
    ENCLAVE_PREVIOUS_PCR0 = "genesis";
    ENCLAVE_TEST_SALT = "blue";
  };

  bluePCR0 =
    lib.toLower
      (builtins.fromJSON
        (builtins.readFile "${blueEif}/pcr.json")).PCR0;

  greenEif = mkEif {
    ENCLAVE_PREVIOUS_PCR0 = bluePCR0;
    ENCLAVE_TEST_SALT = "green";
  };

  greenPCR0 =
    lib.toLower
      (builtins.fromJSON
        (builtins.readFile "${greenEif}/pcr.json")).PCR0;

  /*
   * Enclave VM configuration
   */

  mkEnclaveNode =
    eif:
    { pkgs, ... }:
    {
      imports = [
        (self.lib.mkEnclaveQemuAmi {
          inherit eif vhostDeviceVsock;
          memoryMib = 2048;
          cpuCount = 1;
        })
      ];

      virtualisation = {
        memorySize = 4096;
        cores = 2;
      };

      environment.systemPackages = with pkgs; [
        curl
        jq
        openssl
        self.packages.${pkgs.system}.cli
      ];

      networking.firewall.allowedTCPPorts = [ 443 ];

      systemd.services.mock-imds-forward = {
        description = "Test-VPC IMDS endpoint";
        wantedBy = [ "multi-user.target" ];
        wants = [ "network-online.target" ];
        after = [ "network-online.target" ];

        serviceConfig = {
          Type = "simple";
          ExecStartPre =
            "${pkgs.iproute2}/bin/ip address replace "
            + "169.254.169.254/32 dev lo";
          ExecStart =
            "${pkgs.socat}/bin/socat "
            + "TCP4-LISTEN:80,bind=169.254.169.254,reuseaddr,fork "
            + "TCP4:aws:1338";
          Restart = "on-failure";
        };
      };

      systemd.services.enclave-start = {
        requires = [ "mock-imds-forward.service" ];
        after = [ "mock-imds-forward.service" ];
      };
    };

  /*
   * AWS-services VM configuration
   */

  awsNode =
    { pkgs, nodes, ... }:
    {
      virtualisation = {
        memorySize = 2048;
        cores = 1;
      };

      environment.systemPackages = [
        pkgs.awscli2
        pkgs.curl
        pkgs.jq
        pkgs.openssl
        enclaveTofu
      ];

      environment.variables = {
        AWS_ACCESS_KEY_ID = "000000000000";
        AWS_SECRET_ACCESS_KEY = "test";
        AWS_DEFAULT_REGION = "us-east-1";
        AWS_REQUEST_CHECKSUM_CALCULATION = "when_required";
        AWS_RESPONSE_CHECKSUM_VALIDATION = "when_required";
      };

      networking = {
        hosts."${nodes.green.networking.primaryIPAddress}" = [
          "enclave.test"
        ];

        firewall.allowedTCPPorts = [
          1338
          4000
          4566
          14000
        ];
      };

      environment.etc."pebble/ca.crt".source =
        "${pebbleFixtures}/ca.crt";

      systemd.services = {
        ministack = {
          description = "MiniStack AWS emulator";
          wantedBy = [ "multi-user.target" ];

          environment = {
            BIND_HOST = "0.0.0.0";
            GATEWAY_PORT = "4566";
            MINISTACK_REGION = "us-east-1";
            LOG_LEVEL = "INFO";
          };

          serviceConfig = {
            Type = "simple";
            ExecStart = "${ministack}/bin/ministack";
            Restart = "on-failure";
            StateDirectory = "ministack";
            WorkingDirectory = "/var/lib/ministack";
          };
        };

        awsmocks = {
          description = "Attested KMS proxy and IMDS stub";
          wantedBy = [ "multi-user.target" ];
          wants = [ "ministack.service" ];
          after = [ "ministack.service" ];

          environment = {
            KMS_PROXY_LISTEN_ADDR = ":4000";
            IMDS_LISTEN_ADDR = ":1338";
            UPSTREAM_KMS_URL = "http://127.0.0.1:4566";
          };

          serviceConfig = {
            Type = "simple";
            ExecStart = "${awsmocks}/bin/awsmocks";
            Restart = "on-failure";
          };
        };

        pebble = {
          description = "Pebble ACME test server";
          wantedBy = [ "multi-user.target" ];

          environment = {
            PEBBLE_VA_NOSLEEP = "1";
            PEBBLE_WFE_NONCEREJECT = "0";
          };

          serviceConfig = {
            Type = "simple";
            ExecStart =
              "${pkgs.pebble}/bin/pebble "
              + "-config ${pebbleFixtures}/pebble-config.json "
              + "-strict false";
            Restart = "on-failure";
          };
        };
      };
    };

in

pkgs.testers.runNixOSTest {
  name = "enclave-blue-green-e2e";

  nodes = {
    aws = awsNode;
    blue = mkEnclaveNode blueEif;
    green = mkEnclaveNode greenEif;
  };

  testScript =
    ''
      BLUE_PCR0 = ${builtins.toJSON bluePCR0}
      GREEN_PCR0 = ${builtins.toJSON greenPCR0}
      AWS_NODE_IP = ${builtins.toJSON awsNodeIP}
    ''
    + builtins.readFile ./e2e.py;
}
```

nix/tests/test-app.nix
nix/tests/awsmocks.nix
nix/tests/ministack.nix
nix/tests/pebble.nix

are small enough and uesd one that it could be folded.

Also let us try to fold test-app.nix in defaults.nix, and maybe call it suite:

It basically does this:
Enclave test suite
├── AMI build
├── EIF build
├── OpenTofu validation
└── end-to-end test

nix/mk-enclave-ami.nix
aruokhai: not really our business like we discussed

aruokhai (commented):
I reckon moving away from the framework approach is the right direction, especially since this is only being used for Arkade products. It should reduce deployment complexity and the surface area for deployment-related bugs. The trade-off is that some things previously handled automatically through the CLI now need to be handled explicitly.

That said, this PR currently combines several different scopes:

1. Blue-green deployment testing
2. Rollback-resistant KV storage
3. Nix-only test requirements
4. Supervisor removal

My thoughts on each:

* **3 — Nix-only test requirements:** Looks good to me overall, aside from a few minor file-organisation issues.
* **4 — Supervisor removal:** I think this could have been a separate PR. The rationale of reducing code and maintenance burden is fair, but combining it here makes the overall change harder to review.
* **2 — Rollback-resistant KV:** Seems reasonable to remove for now. A design like this needs to be properly tested and reviewed before we can be confident that its security properties actually hold.
* **1 — Blue-green deployment testing:** The test builds the AMI but does not actually test the resulting AMI. Instead, it extracts the host-specific functionality and enclave setup into a reusable module and uses that module for the blue-green test. I think the scope of the test itself is adequate, but we should be clear that this is testing the deployment logic rather than the final AMI artifact end-to-end.

aruokhai:
@arkana-ai-bot  remember, this PR doesn't add new functionalities

chris-ricketts:
@aruokhai nix surface area has been reduced to the following plus the root `flake.nix`:

```
❯ rg --files nix
nix/tests/default.nix
nix/tests/e2e.py
nix/tests/awsmocks/main.go
nix/tests/awsmocks/go.sum
nix/tests/awsmocks/go.mod
nix/tests/test-app/main.go
nix/tests/test-app/go.mod
nix/build-eif.nix
```

s373nZ:
What do you think about decoupling the EIF file from the AMI in the Nix builder, as, say, a S3 or general (remote) file location? I am exploring to try and build a real AMI using the Nix tooling as a first step before re-implementing everything in Ansible.

Also -- apparently, if we were to build an AMI with Nix, the image inherits an `amazon-init.service` unit which runs on *every* boot.  It reads /etc/ec2-metadata/user-data and does one of two things:
1. Starts with `#!`? Then runs it as a bash script.
2. Anything else non-empty? Then treats it as a NixOS configuration, copies it to `/etc/nixos/configuration.nix`, and rebuilds the whole system.

Seems like this is the Nix version of the AMI user-data script (although that script usually runs only on initial boot iirc). It could be leveraged, but AI research suggests it might be better to place the EIF loading in a custom unit and disable `amazon-init` for security reasons if we're not using it....

If this is a direction we want to go, does it even belong in this PR?

aruokhai:
@s373nZ  I think we have cleaned up the PR to only include NIX files necessary for testing the runtime, and also a single buildEIF file, for building the EIF file. Therefore, all magic that involves building the AMI does not belong in the codebase or PR.

.github/workflows/integration-test.yml
```diff
@@ -7,103 +7,24 @@ on:
...
-          cp /tmp/pcr-v3.json .enclave/artifacts/pcr-v3.json
-          cp /tmp/image-v4.eif .enclave/artifacts/image-v4.eif
-          cp /tmp/pcr-v4.json .enclave/artifacts/pcr-v4.json
-
       - name: Run integration test
-        run: cd test && docker compose --profile test run --build test-runner
-        env:
-          BOOT_TIMEOUT: "120"
-          INIT_TIMEOUT: "300"
+        run: nix build .#checks.x86_64-linux.e2e --print-build-logs
```
s373nZ: Does this need to be hardcoded to `x86_64` to run? Would we be able to run the tests from an ARM machine?

aruokhai: yes, there is config in the nix root file I believe  tthat include necessary packages  for both arm and x86

chris-ricketts:
@arkana-ai-bot 

Finding 1: now tracked in #157 as it was already present on `master`

Findings 2, 3, and 7 concern the per-boot response-signing mechanism. PR #151 removes that mechanism entirely because per-node signing keys are incompatible with horizontally scaled enclaves behind a load balancer. user_data intentionally becomes the 39-byte TLS certificate hash, and client authentication relies on attested TLS pinning. We will not add validation or compatibility code to a path scheduled for deletion.

Finding 4: the host/operator is trusted to call migration endpoints

Finding 5: `enclave-skeleton` is an obsolete repo that will be archived

Finding 6: removed in 0039dc1d66175fc9ce217beb00ccaf18e3c1442f

N2: removed in ec4021f8d78aa4d30bc4a37d74134bb82f9b9ba3

.github/workflows/integration-test.yml
chris-ricketts: In theory it works for both `aarch64-linux` and `x86_64-linux` systems, but so far only tested on `x86_64` 

I will create an issue to add an `aarch64` integration test target in CI

s373nZ:
Tested by building this branch on an actual AWS ARM instance and ran into the following issues described by Claude:

---

Evaluating this branch for a downstream Ansible/Packer deployment — built and booted an EIF from it on real Nitro hardware. The Nix API refactor made that straightforward, and the "any deployment system may be used" framing in the Deployment section is exactly the seam we needed. Four findings, reviewed at `ae146e8`.

**1. aarch64 is listed as supported but cannot boot** — filed as #158.

An aarch64 EIF builds and measures correctly (`describe-eif` on Graviton matched `pcr.json` byte for byte), but the runtime dies at clock sync: `open /dev/ptp0: no such file or directory`. Root cause is AWS's kernel blob, not this repo — `blobs/aarch64/Image.config` has no `CONFIG_PTP_1588_CLOCK_KVM` while `blobs/x86_64/bzImage.config` sets it `=y`, and both are `4.14.256-209.484.amzn2`. Not fixable by bumping the blob source; the aarch64 `Image` is unchanged between nitro-cli `v1.2.3` and `v1.4.5`. Details and reproduction in the issue.

**2. `checks.aarch64-linux.e2e` exists but cannot run.** The `optionalAttrs (system == "x86_64-linux")` guard around `e2e` was dropped in this branch, but `nix/tests/default.nix` still hardcodes `qemu-system-x86_64`. `qemu_test` is host-CPU-only (`--target-list=i386-softmmu,x86_64-softmmu` on x86_64), so that binary doesn't exist on aarch64 and `nix flake check` fails there. This is also why finding 1 went unnoticed — `e2e` is the only check that boots a real EIF.

**3. The README Quickstart produces an unbootable EIF.** Its `env` block omits `ENCLAVE_MIGRATION_INTENT_RETENTION`, which `validateEnvironment` requires:

```
ERROR runtime failed error="invalid config: ENCLAVE_MIGRATION_INTENT_RETENTION must not be empty"
```

I hit this by copying the Quickstart verbatim. The Runtime configuration table also lists a default of `87600h` for it, but `migrationIntentRetention()` returns an error on empty rather than defaulting — so the documented default doesn't exist. The PR description's example flake does include the variable; it's just the README that doesn't.

**4. README and code disagree on the dev/COSE switch.** The Identity and security table says `ENCLAVE_DEPLOYMENT` defaults to `dev` and that `dev` disables COSE signature verification. In `runtime/environment.go`, `ENCLAVE_DEPLOYMENT` has no default (`validateEnvironment` errors if empty), and `skipCOSEVerification()` keys off `IsDev()`, which reads a separate `ENCLAVE_DEV` variable that the table doesn't document. Flagging this one specifically because it's a security note: someone setting `ENCLAVE_DEPLOYMENT=prod` would reasonably believe verification is on and would never learn about `ENCLAVE_DEV`.

Happy to open PRs for any of 2–4 if useful.

chris-ricketts:
@s373nZ cheers for the review!

1. Nice one for testing out `aarch64` and filing the issue

2. As suggested in #158, I'll drop `aarch64` from the supported enclave systems in this PR.

3: good spot - `buildEif` is pretty simple at this stage, its my intention to make some QoL improvements like a typed environment argument so that it won't compile if the required params are not provided, but this is fairly low priority for now. I'll add a commit to update the README in this PR.

4. I'll push a commit to align the README with the code.

s373nZ:
@chris-ricketts I was really hoping we can still run on aarch64 for Graviton instances - cheaper and more efficient. While the issue suggests to remove support, I was hoping there might be a fix to the runtime or the clocksync work that could facilitate both. Let's chat about it tomorrow before we rip ARM out entirely?

chris-ricketts:
@s373nZ agree `aarch64` support is desirable 

In 8aa973d3f9429c36a4f8798f3c6a64f4742c620b I just did a one line removal in `flake.nix` and some README updates.

It'll be easy enough to add back in a new PR that properly satisfies the kernel requirements described in #158

PR #151, opened 2026-08-12
Horizontal scaling

aruokhai:
<html><head></head><body><h1>Horizontal Scaling</h1>
<p>Run N identical-PCR0 enclaves as a shared fleet, using the same IAM role, KMS key, SSM tree, storage bucket, KV table, and TLS certificate.</p>
<h2>Status</h2>

Area | Status
-- | --
Runtime correctness at N>1 | Done — genesis, boot gate, response signing
Fleet-shared certificate (DNS-01) | Done
Infrastructure (count, NLB, private subnet, NAT) |  done — OpenTofu still deploys one instance
Fleet migration | Out of scope

<p>The runtime is safe at N&gt;1 today; <code>make test-fleet</code> proves this with two concurrent enclaves. Real AWS deployment still requires the infrastructure work below.</p>
<h2>Shared State</h2>
<p>Same-PCR0 enclaves derive the same DEK from the same KMS key, allowing any node to decrypt data sealed by another.</p>
<ul>
<li><strong>KMS:</strong> policy gates on <code>RecipientAttestation:PCR0</code>, not instance identity.</li>
<li><strong>SSM:</strong> shared <code>/{deployment}/{app}/…</code> tree, including the genesis <code>KMSKeyID</code>.</li>
<li><strong>S3:</strong> sealed certificate, ACME account key, and mutex objects.</li>
<li><strong>DynamoDB:</strong> shared version vector and freshness anchor.</li>
</ul>
<h2>Genesis</h2>
<p>Genesis is serialized with an S3 lease before <code>CreateKey</code>. This prevents concurrent cold boots from creating multiple keys and committing the wrong one.</p>
<pre><code class="language-text">acquire "genesis" lease (bounded to 10m)
  → re-read KMSKeyID
      set?   → release, resume
      unset? → create key, seal data, commit KMSKeyID, release
</code></pre>
<p>The lease uses S3 conditional writes and an in-memory ETag as the claim. Heartbeats run at TTL/3; <code>412</code> means the lease was lost. <code>409</code> is a retryable S3 race, not contention.</p>
<p>Because the SSM commit is unconditional, <code>lease.Verify()</code> runs immediately before it to prevent an expired/descheduled holder from clobbering a newer owner.</p>
<p>The lease is not a security boundary: the instance role can manipulate it. Compromise can cause denial of renewal, but cannot mint or read a certificate without the required external permissions and PCR0-bound decryption.</p>
<h2>Fleet Certificate</h2>
<p><code>tls-alpn-01</code> cannot work reliably behind a load balancer, so the fleet uses DNS-01. ACME-enabled enclaves require a Route53 hosted zone.</p>
<p>The certificate and key are stored together in one DEK-sealed S3 object. Renewal is serialized by an <code>acme-renewal</code> lease and performed proactively:</p>
<pre><code class="language-text">poll every 5–15m with jitter
  → detect certificate ETag change and adopt peer renewal
  → &lt;30d remaining: acquire lease and renew
  → &lt;7d + stale: log loudly + metric
</code></pre>
<p>Three invariants are critical:</p>
<ol>
<li>Capture the certificate ETag <strong>before</strong> starting an order and never refresh it before the write.</li>
<li>Write to S3 before updating in-memory certificate state.</li>
<li>On <code>412</code>, reload the peer's certificate; <strong>never retry the ACME order</strong>.</li>
</ol>
<p>The certificate and attested hash are installed together through one <code>atomic.Pointer</code>, preventing the node from serving one leaf while attesting another.</p>
<h2>Boot Gate</h2>
<p><code>freshnessAnchor.Establish</code> now re-reads a stale KV key with a strongly consistent <code>GetItem</code> before halting. This avoids bricking a joining node when a peer commits between the initial version-vector scan and anchor listing. Failed re-reads do not halt boot.</p>
<h2>Response Signing</h2>
<p>The per-boot response-signing key is removed. It caused verification failures behind a load balancer because each enclave had a different key.</p>
<p>Authenticity now comes from attested TLS: clients pin the live certificate against <code>tlsKeyHash</code>. <code>user_data</code> is now 39 bytes: <code>"sha256:"</code> plus the raw 32-byte digest.</p>

<h2>Testing</h2>
<p><code>make test-fleet</code> boots two enclaves from the same EIF, with distinct vsock ports supplied through the kernel command line.</p>
<p>It verifies:</p>
<ul>
<li>Shared genesis and resume behavior.</li>
<li>Exactly one fleet certificate.</li>
<li>No duplicate ACME orders or certificate writes.</li>
<li>Cross-node decryption of sealed data.</li>
<li>Certificate attestation on both nodes.</li>
<li>Survival after node failure.</li>
<li>Endpoint parity and concurrent request handling.</li>
<li>Real certificate pinning.</li>
</ul>
<p>The strongest checks are B reading data sealed by A, proving shared DEK derivation, and <code>/test/pcr-secrets</code> returning the same value through both nodes, proving both decrypted the same ciphertext with the same KMS key.</p>
<p><code>test-acme</code> and <code>test-fleet</code> share <code>test/app/tofu</code> and must not run concurrently because teardown can delete <code>.terraform</code> during the other test.</p>
<h3>Not Covered</h3>
<p>Large-N single issuance and renewal contention are covered at unit level with <code>fakeS3</code>, not end-to-end. Validate both on a two-instance staging deployment before merging the infrastructure changes.</p></body></html>


aruokhai:
**M3 — won't fix.**

The finding assumes `time.After` timers aren't GC'd until they fire. That was true before **Go 1.23**; since then, unreferenced timers are immediately eligible for GC. This module targets **Go 1.25.0**, so there is no leak.

The proposed accumulation also can't occur here. The only `select` cases are the timer and `ctx.Done()`, and `ctx.Done()` **returns from the function**. Thus, the timer must fire before the next iteration: at most one timer is live per goroutine, with timers created strictly sequentially. The only abandoned timer is when cancellation wins, which exits the loop—one per call, not one per poll.

The "boot storm with N enclaves" doesn't compound this: each enclave is a separate process/VM, and `awaitGenesis` uses one goroutine, so N enclaves means N processes with one timer each.

Using `NewTimer`/`Stop`/`Reset` adds mutable state and several lines without freeing anything.

runtime/devcmdline.go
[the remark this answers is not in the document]
aruokhai: yes, we should remove

runtime/lease.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+}
+
+func (l *Lease) Context() context.Context { return l.ctx }
+
+func (l *Lease) Verify(ctx context.Context) error {
+	if l.ctx.Err() != nil {
+		return fmt.Errorf("%w: %s: %w", ErrLeaseLost, l.key, context.Cause(l.ctx))
+	}
+	_, live, err := l.get(ctx)
```
chris-ricketts: `Verify` discards lease document, so expired lease can still pass. After repeated heartbeat failures, `l.ctx` may remain live past `expires_at` because expiry is checked only after heartbeat attempts. Same ETag then returns success while peer may steal. Should `heartbeat` cancel on an expiry timer and `Verify` reject `doc.lapsed(time.Now())`?

https://github.com/ArkLabsHQ/enclave/blob/8a25bdff7ab41e9dab6aa37df23d8623a7dad0ee/runtime/lease.go#L177-L190

Expiry is checked here, but only after a failed renewal returns. If a failure occurs before expiry, default waits until the next `ttl/3` tick but the lease can expire during that wait or the next in-flight PUT while `l.ctx` remains live.

runtime/lease.go
```diff
@@ -0,0 +1,382 @@
...
+		case l.lapsed():
+			l.cancel(
+				fmt.Errorf("%w: %s: heartbeat failed past expiry: %w", ErrLeaseLost, l.key, err),
+			)
+			return
+		default:
+			slog.Warn("lease heartbeat failed, retrying", "key", l.key, "error", err)
+		}
+	}
+}
```
chris-ricketts: Following up on [this comment](https://github.com/ArkLabsHQ/enclave/pull/151/changes#r3871427347), this keeps existing renewal cadence but wakes the heartbeat at local expiry and prevents an in-flight PUT from exceeding remaining lease lifetime.

Precision note: local `expiresAt` includes subsecond precision, while stored `expires_at` is truncated to seconds. A peer may therefore consider the lease stealable up to one second earlier. With the proposed `Verify` expiry check, the safety-sensitive genesis commit remains protected; subsecond context lag seems acceptable for a five-minute TTL.

```suggestion
func (l *Lease) heartbeat() {
	defer close(l.done)

	ticker := time.NewTicker(l.ttl / 3)
	defer ticker.Stop()

	for {
		l.mu.Lock()
		expiresAt := l.expiresAt
		l.mu.Unlock()

		select {
		case <-l.stop:
			return
		case <-time.After(time.Until(expiresAt)):
			l.cancel(fmt.Errorf("%w: %s: expired", ErrLeaseLost, l.key))
			return
		case <-ticker.C:
		}

		ctx, cancel := context.WithTimeout(
			context.Background(),
			min(leaseWriteTimeout, time.Until(expiresAt)),
		)
		guard, nextExpiry, err := l.put(ctx)
		cancel()

		switch {
		case err == nil:
			l.renewedAt(guard, nextExpiry)
		case isPreconditionFailed(err):
			l.cancel(fmt.Errorf("%w: %s", ErrLeaseLost, l.key))
			return
		case l.lapsed():
			l.cancel(
				fmt.Errorf("%w: %s: heartbeat failed past expiry: %w", ErrLeaseLost, l.key, err),
			)
			return
		default:
			slog.Warn("lease heartbeat failed, retrying", "key", l.key, "error", err)
		}
	}
}
```

runtime/lease.go
```diff
@@ -0,0 +1,382 @@
...
+			return false, nil
+		}
+
+		l.setGuard(theirs)
+		guard, expiresAt, err = l.put(ctx)
+		switch {
+		case err == nil:
+			l.guard, l.expiresAt = guard, expiresAt
+			return true, nil
+		case isPreconditionFailed(err):
```
chris-ricketts: [S3 documents](https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html#conditional-error-response) that an `If-Match` PUT can return `404` when a concurrent delete wins.

The same status needs handling in the heartbeat:

https://github.com/ArkLabsHQ/enclave/blob/8a25bdff7ab41e9dab6aa37df23d8623a7dad0ee/runtime/lease.go#L177-L190

For heartbeat renewal, `404` means the lease object is gone, so ownership is definitely lost and it should be handled like `412`. 

In `claim`, during stealing, `404` means the old holder released between GET and PUT, so claim should retry its create-only path. Currently heartbeat remains live until expiry, while claim returns a fatal error.

Could both paths explicitly handle `isNoSuchKey(err)` and get targeted tests? `fakeS3` currently maps missing `If-Match` to `412`, so it will not reproduce real S3 behavior.

runtime/acme_issuer.go
```diff
@@ -0,0 +1,309 @@
...
+
+	keyPEM, err = encodeECKey(key)
+	if err != nil {
+		return nil, nil, err
+	}
+	return encodeCertChain(chain), keyPEM, nil
+}
+
+// register creates the ACME account. The account key is fleet-shared, so on
+// every enclave but the first this is a no-op.
```
chris-ricketts: Nit: explain the no-op more clearly and where the account key comes into it

```suggestion
// register creates or discovers the fleet's ACME account using i.client.Key.
// Later enclaves reuse the existing account associated with the shared key.
```

runtime/tls.go
```diff
@@ -121,40 +190,9 @@ func loadTLSConfigOverridesFromSSM(ctx context.Context, ssm SSM, cfg *Config) er
...
-			}
-		}
-		return cert, err
-	}, nil
-}
-
 // configureSelfSignedCert generates an ECDSA-P256 leaf cert, records its fingerprint
 // in the attestation hashes so clients can pin it against the NSM document
-func configureSelfSignedCert(cfg *Config, hashes AttestationHashes) (TLSCertCallback, error) {
+func configureSelfSignedCert(cfg *Config, hashes *AttestationHashes) (TLSCertCallback, error) {
```
chris-ricketts: Why not also create the self-signed cert under a lease and share it across the fleet?

runtime/certmanager.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("acquire renewal lease: %w", err)
+	}
+	if lease == nil {
+		return nil, nil // a peer is renewing
+	}
+	defer func() { _ = lease.Release(context.WithoutCancel(ctx)) }()
+
+	return m.renew(ctx, certETag)
+}
```
chris-ricketts: [`resolve` already uses safer ordering](https://github.com/ArkLabsHQ/enclave/blob/2f6b33041108579fcb7b365cefd05e9cf27f3d4a/runtime/certmanager.go#L95-L109): acquire lease, reload stored cert, then issue only if still needed. Renewal skips that reload, so a peer can commit and release between tick’s read and our acquisition, causing an unnecessary ACME order. Renewal also keeps using caller context after heartbeat detects lease loss.

Could we extract held-lease work into renewWithLease? It should cancel on caller or lease context, reload under lease, return stored bundle only when not due, otherwise renew against latest ETag. resolve can then call it after blocking acquisition with `m.renewWithLease(ctx, lease, "")`.


```suggestion
func (m *certManager) renewUnderLease(ctx context.Context, certETag string) (*certBundle, error) {
	lease, err := TryAcquireLease(ctx, m.s3, m.leaseBucket, certLeaseName, leaseTTL)
	if err != nil {
		return nil, fmt.Errorf("acquire renewal lease: %w", err)
	}
	if lease == nil {
		return nil, nil // a peer is renewing
	}
	defer func() { _ = lease.Release(context.WithoutCancel(ctx)) }()

	return m.renewWithLease(ctx, lease, certETag)
}

func (m *certManager) renewWithLease(
	ctx context.Context,
	lease *Lease,
	certETag string,
) (*certBundle, error) {
	renewCtx, cancel := context.WithCancelCause(ctx)
	stop := context.AfterFunc(lease.Context(), func() {
		cancel(context.Cause(lease.Context()))
	})
	defer cancel(nil)
	defer stop()

	if cause := context.Cause(lease.Context()); cause != nil {
		cancel(cause)
	}

	bundle, err := m.store.LoadCert(renewCtx)
	if err != nil {
		return nil, err
	}
	if bundle != nil {
		if time.Until(bundle.notAfter) > renewBefore {
			return bundle, nil
		}
		certETag = bundle.etag
	}
	return m.renew(renewCtx, certETag)
}
```

runtime/acme_issuer.go
```diff
@@ -0,0 +1,309 @@
...
+		return nil, nil, fmt.Errorf("create CSR: %w", err)
+	}
+
+	chain, _, err := i.client.CreateOrderCert(ctx, order.FinalizeURL, csr, true)
+	if err != nil {
+		return nil, nil, fmt.Errorf("finalize order for %q: %w", domain, err)
+	}
+	if len(chain) == 0 {
+		return nil, nil, errors.New("CA returned an empty certificate chain")
+	}
```
chris-ricketts: `CreateOrderCert` only PEM-decodes the response and explicitly leaves expected-feature validation to callers. `SaveCert` checks the leaf/key pair, but `tls.X509KeyPair` ASN.1-parses only the leaf and does not check its hostname or validity window. Could we parse each returned DER certificate and, before persisting it, reject a leaf for which `VerifyHostname(domain)` fails or the current time is outside `[NotBefore, NotAfter]`? This mirrors the relevant checks in `autocert.validCert` and prevents a buggy/private CA response from becoming the fleet-wide served certificate.

`CreateOrderCert` comment from `x/crypto/acme/rfc8555.go`:
```
// Callers are encouraged to parse the returned value to ensure the certificate is valid and has the expected features.
```

runtime/migration_intent.go
```diff
@@ -109,6 +114,74 @@ func (l *migrationIntentLog) Head(ctx context.Context) (*migrationIntent, error)
...
+				errMigrationIntentStoreUnavailable,
+			)
+		}
+		keyMarker, versionMarker = out.NextKeyMarker, out.NextVersionIdMarker
+	}
+}
+
+// Genesis records that this enclave created the deployment, so no later boot can
+// claim genesis again.
+func (l *migrationIntentLog) Genesis(ctx context.Context) (*migrationIntent, error) {
```
chris-ricketts: Could genesis be a dedicated Object-Locked marker key instead of a migration action? It represents one durable boolean and would avoid adding genesis-specific cases to the migration state machine.

chris-ricketts (changes requested):

runtime/boot.go
```diff
@@ -0,0 +1,749 @@
...
+	return pcr0, attestation, nil
+}
+
+func (b *Boot) committedKeyID(ctx context.Context) (string, error) {
+	keyID, err := b.ssm.MayGet(ctx, kmsKeyIDParam())
+	if err != nil {
+		return "", fmt.Errorf("failed to get KMS key ID SSM param: %w", err)
+	}
+	return keyID, nil
+}
```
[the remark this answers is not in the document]
aruokhai: 7a837c2b182be76b8c6c4aff6298607cce094afd

runtime/boot.go
```diff
@@ -0,0 +1,749 @@
...
+	state := &planned.state
+	kms, err := FetchOrCreatePrimaryKMS(
+		ctx, b.nsm, b.kmsAPI, b.sts,
+		state.kmsKeyID, state.predecessorPCR0, state.predecessorAttestation,
+	)
+	if err != nil {
+		return bootResult{}, fmt.Errorf("failed to fetch/create primary KMS key: %w", err)
+	}
+	return b.establish(ctx, planned, kms)
+}
```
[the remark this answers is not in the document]
aruokhai: 7a837c2b182be76b8c6c4aff6298607cce094afd

runtime/boot.go
```diff
@@ -0,0 +1,749 @@
...
+) error {
+	if err := writeOriginReceipt(ctx, nsm, ssm, kms, snapshotRoot, state.currentPCR0); err != nil {
+		return err
+	}
+
+	if b.lease != nil {
+		if err := b.lease.Verify(ctx); err != nil {
+			return fmt.Errorf("refusing to commit genesis: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
aruokhai: 63cc0c99f3c9e7452d6163d0cbba2ae030893f30

runtime/boot.go
```diff
@@ -0,0 +1,749 @@
...
+	if err := ssm.Set(ctx, kmsKeyIDParam(), kms.KeyID()); err != nil {
+		return fmt.Errorf("failed to commit genesis KMS key ID: %w", err)
+	}
```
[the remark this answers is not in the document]
aruokhai: 63cc0c99f3c9e7452d6163d0cbba2ae030893f30

runtime/migration_intent.go
aruokhai: 63cc0c99f3c9e7452d6163d0cbba2ae030893f30

runtime/boot.go
```diff
@@ -0,0 +1,749 @@
...
+	if err := planned.mode.verifySnapshot(state, b.nsm, snapshotRoot); err != nil {
+		return bootResult{}, err
+	}
+
+	if kms.KeyID() != snapshot.kmsKeyID {
+		return bootResult{}, fmt.Errorf(
+			"state KMS key ID %q does not match active KMS key ID %q",
+			snapshot.kmsKeyID, kms.KeyID(),
+		)
+	}
```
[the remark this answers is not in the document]
aruokhai: 10f9011107b9b7136ea3b4ef1bf6dc33ab021d5e

runtime/boot.go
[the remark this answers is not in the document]
aruokhai: 63cc0c99f3c9e7452d6163d0cbba2ae030893f30

runtime/boot.go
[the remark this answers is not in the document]
aruokhai: 63cc0c99f3c9e7452d6163d0cbba2ae030893f30

runtime/boot.go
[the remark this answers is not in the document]
aruokhai: 10f9011107b9b7136ea3b4ef1bf6dc33ab021d5e

runtime/lease.go
aruokhai: ade052ca79c2e83e57ac3bf29deee14a52d40bad

runtime/lease.go
aruokhai: 3c3dec1893fdb8d5959e3bf1caa29b4bd79097fc

runtime/certmanager.go
aruokhai: 7886fa0e7bdd657a7a99e241c16fd62a59b09d4e

runtime/acme_issuer.go
aruokhai: b2f775f4316beb145de328ae49cd84888ce72c7b

aruokhai:
@arkana-ai-bot — R2 and U2 are fixed in the working tree and land in the next push. Notes on the rest.

**R2 — fixed.** Both paths now log: `tick()` on adopting a peer's renewal, and `renew()` on a successful issuance, each with the FQDN and `not_after`. Worth recording why this is more than a log line: rotation changes the attested TLS hash, so every client holding a pinned hash from bootstrap starts failing until it re-bootstraps. Without a rotation event there is nothing to correlate those failures against.

**U2 — fixed, and it was more than a test-fidelity gap.** Making `fakeS3` honour context cancellation showed the production behaviour was wrong, not merely untested: with a cancelled `renewCtx`, `LoadCert` fails first and callers received a bare `context.Canceled` with no indication that the lease was the reason. `renewWithLease` now wraps store and issuer errors with `context.Cause`, so a lost lease surfaces as `ErrLeaseLost` regardless of which call noticed first. All five `fakeS3` methods check `ctx.Err()`, and the existing test now exercises the production path — it fails without the wrapping.

**S1 — already addressed on this branch.** `determineMode` assigns `state.kmsKeyID = keyID` at `runtime/boot.go:227`, *above* the `genesisArtifact == nil` branch, so `genesisBoot.verify` rejects an interrupted genesis during `plan()` with "genesis boot with a committed KMS key" — before `Boot` reaches `FetchOrCreatePrimaryKMS`. No key is minted and no receipt is written. `TestGenesisRequiresAnEmptyGenesisIntent/a_committed_key_with_an_empty_log_is_a_wiped_log` asserts exactly that, including `require.Empty(t, fx.kmsf.keys)` at `runtime/boot_test.go:1028`.

**M3 and T1 — not accumulating.** `time.After` allocates a timer, not a goroutine, and since Go 1.23 an unreferenced timer is eligible for collection immediately whether or not it has fired; `runtime/go.mod` is `go 1.25.0`. Both loops also block on the select, so at most one timer is outstanding per iteration. Leaving both as they are.

**U1 — acknowledged, deliberate.** The HEAD-first check went away with the `tick()` redesign. Agreed it is negligible at current N; we will revisit if the fleet grows.

**R3 — noted.**

runtime/migration_intent.go
aruokhai: yes, that seems to be more ideal

README.md
```diff
@@ -365,7 +373,6 @@ With `D` = deployment, `A` = app name, `L` = `locked` or `unlocked`:
 | Path | Written by | Purpose |
 |---|---|---|
 | `/D/A/TLSCacheBucketName` | operator | ACME certificate cache bucket. |
```
s373nZ: Is `TLSCacheBucketName` replaced by `CertBucketName` and `LeaseBucketName`?

README.md
```diff
@@ -448,22 +455,28 @@ interfaces.
 
 ### AWS requirements
 
-Create a private S3 bucket for the ACME cache and a separate private S3 bucket
-for migration intents. The intent bucket must have versioning and Object Lock
-enabled when it is created; the runtime applies retention to each intent object.
-Write their names to these SSM parameters:
+Create a private S3 bucket for the ACME cache and write its name to
+`/<deployment>/<app>/TLSCacheBucketName`.
```
s373nZ: Same... Is `TLSCacheBucketName` replaced by `CertBucketName` and `LeaseBucketName`?

s373nZ (commented):
Two general questions:
1. What is port `6379` in the PR description used for? It that still relevant?
2. I looks like the leases in the `LeaseBucket` don't need to be S3 object versioned... Should be ok to create that bucket without versioning, right?

With regard to the horizontal scaling from the infrastructure side, my assumption right now is to have a blue ASG and a green ASG and horizontally scaling them out should be tantamount to bumping the `desired_count` up for a given group.

runtime/boot.go
```diff
@@ -0,0 +1,813 @@
...
+		return fmt.Errorf("refusing to commit genesis without lease")
+	}
+
+	if err := writeOriginReceipt(ctx, nsm, ssm, kms, snapshotRoot, state.currentPCR0); err != nil {
+		return err
+	}
+
+	if err := b.lease.Verify(ctx); err != nil {
+		return fmt.Errorf("refusing to commit genesis: %w", err)
+	}
```
chris-ricketts: Verify lease before any side effects

```suggestion
	if err := b.lease.Verify(ctx); err != nil {
		return fmt.Errorf("refusing to commit genesis: %w", err)
	}

	if err := writeOriginReceipt(ctx, nsm, ssm, kms, snapshotRoot, state.currentPCR0); err != nil {
		return err
	}

```

Additionally, should `writeOriginReceipt` also use `Overwrite=False` for the same reason as the KMS ID?

README.md
aruokhai: 1- 6379 is not necessary, It was necessary when the e2e test included  NLB and ASG, but not needed anymore
2 - Yes, that is true, scaling is not about  correctness, the host can refuse to scale, but I dont think it hurts here

README.md
aruokhai: @s373nZ  I thought we wanted to scale manually, Should be good anyways,  to validate with ASG in production. Probably good to try it with this branch

runtime/genesis.go
```diff
@@ -0,0 +1,239 @@
...
+		BucketName: l.bucket,
+		PCR0:       pcr0,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("encode deployment genesis: %w", err)
+	}
+	doc, _, err := l.nsm.BuildAttestationDocument(WithUserData(payload))
+	if err != nil {
+		return nil, fmt.Errorf("attest deployment genesis: %w", err)
+	}
```
chris-ricketts: Who is the intended verifier of the attestation document? It is not currently verified by the runtime itself.

README.md
s373nZ: Changing `desired_count` is still a manual process IMO. We're not setting up autoscaling parameters or anything.

runtime/boot.go
aruokhai: `writeOriginReceipt` writes to a path scoped by the genesis attempt’s KMS key ID, so competing genesis attempts do not overwrite each other’s receipts. If an enclave loses the lease after writing its receipt, that receipt remains an unselected orphan because the enclave never claims the shared KMSKeyID.
writeOriginReceipt performs NSM work and an SSM request, which introduces latency between operations. Moving the only lease verification before it would increase the interval between verification and the critical shared-state commit.
Keeping Verify() after writeOriginReceipt places it immediately before the create-only KMSKeyID write. This protects the critical operations that select and finalize the winning genesis state.

runtime/genesis.go
aruokhai: It is an artifact that is expected to be verified out of band

README.md
aruokhai: Okay, nice

aruokhai:
@arkana-ai-bot 
AA1 — parseCertBundle marshals the signer to PKCS#8
We are leaving this unchanged.
tls.X509KeyPair does more than parse PEM: it verifies that the certificate and private key have matching types and public components. For ECDSA, it explicitly compares the public-key coordinates:

> if pub.X.Cmp(priv.X) != 0 || pub.Y.Cmp(priv.Y) != 0 {
>     return fail(errors.New("tls: private key does not match public key"))
> }

Direct construction would move this validation into our code. The current key is a software P-256 key and is already serialized during genesis and migration, so non-exportable signers are not supported by the wider design. This is not a hot path, making the serialization overhead negligible. We prefer the standard-library validation.
BB1 — non-enclave fallback removed
Correct and intentional.
This runtime is being prepared for fleet deployment, where every node must use the persistent deployment TLS key and shared certificate storage. Silently generating a per-process certificate outside an enclave would violate that invariant. QEMU tests provide the same required SSM parameters and buckets as the production path.

aruokhai:
@arkana-ai-bot V2 applies only to the test environment. We prefer to fail fast here rather than return a successful response to the runtime when the TXT record was nt synchronized to Pebble’s DNS server. Catching and suppressing the exception would hide the actual failure and cause ACME validation to fail or time out later.

chris-ricketts (approved):
LGTM

PR #159, opened 2026-08-25
Single PCR scoped KMS_ID

aruokhai:
## Summary

This PR changes blue/green migration so every enclave generation retains its own KMS key, ciphertexts, and commit pointer.

Instead of replacing a shared `KMSKeyID`, commit pointers and predecessor artifacts are scoped by PCR0. The predecessor remains healthy and restartable after preparing its successor, eliminating the need for rollback-specific runtime machinery.

## Changes

- Scope KMS commit pointers as `KMSKeyID/<pcr0>`.
- Scope predecessor PCR0 and attestation parameters by successor PCR0.
- Create migration keys that admit only the successor’s PCR0.
- Preserve the predecessor’s key, ciphertexts, and commit pointer.
- Reject self-targeted migrations.
- Reject finalisation when the target already has a commit pointer.
- Remove the rollback halt circuit breaker and dual-PCR key handling.
- Update state-root receipts to include the PCR0-scoped commit path.
- Extend unit and E2E coverage to verify that:
  - the successor can adopt migrated state;
  - the predecessor cannot decrypt the successor’s migration key;
  - the predecessor remains healthy after handoff;
  - the predecessor can restart using its original state;
  - duplicate finalisation is rejected.
- Update documentation for the new state and migration model.

## New migration model

```text
Predecessor PCR0
├── original KMS key
├── original ciphertexts
└── KMSKeyID/<predecessor PCR0>

Successor PCR0
├── successor-only KMS key
├── re-encrypted ciphertexts
└── KMSKeyID/<successor PCR0>

<!-- codesmith:footer -->
---
<a href="https://app.blacksmith.sh/ArkLabsHQ/codesmith/enclave/pr/159"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg"><img alt="View with [code]smith" src="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"></picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790269899&installation_model_id=433183&pr_number=159&repository=ArkLabsHQ%2Fenclave&return_to=https%3A%2F%2Fgithub.com%2FArkLabsHQ%2Fenclave%2Fpull%2F159&signature=0921a8b7ff58c06e124ed276cdb0b7b8d87388c83ce5468972e5587f0481340d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg"><img alt="Autofix with [code]smith" src="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"></picture></a>

<!-- codesmith:autofix:disabled -->
<!-- /codesmith:footer -->

### CODE REVIEW

File: runtime/boot.go [hunk 1]
```diff
@@ -370,51 +411,62 @@ func (b *Boot) establish(
...
+		return "", "", "", fmt.Errorf("failed to get predecessor attestation SSM param: %w", err)
+	}
+	if (pcr0 != "") != (keyID != "") || (pcr0 != "") != (attestation != "") {
+		return "", "", "", fmt.Errorf(
+			"inconsistent migration predecessor artifacts (pcr0 present=%v, key present=%v, attestation present=%v)",
 			pcr0 != "",
+			keyID != "",
 			attestation != "",
 		)
 	}
```
chris-ricketts: Nit: hard to read condition

```suggestion
	allPresent := pcr0 != "" && keyID != "" && attestation != ""
	allAbsent := pcr0 == "" && keyID == "" && attestation == ""
	if !allPresent && !allAbsent {
		return "", "", "", fmt.Errorf(
			"inconsistent migration predecessor artifacts (pcr0 present=%v, key present=%v, attestation present=%v)",
			pcr0 != "",
			keyID != "",
			attestation != "",
		)
	}
```

runtime/boot.go
```diff
@@ -621,64 +687,63 @@ func (b *resumeBoot) commitSnapshot(
...
-	if state.predecessorPCR0 == "" {
-		return fmt.Errorf("previous PCR0 is required")
-	}
-	if state.predecessorAttestation == "" {
-		return fmt.Errorf("previous PCR0 attestation is required")
-	}
-	if !strings.EqualFold(eifPreviousPCR0, state.predecessorPCR0) && !isRollbackToSelf(state) {
-		return fmt.Errorf(
-			"previous PCR0 SSM param does not match previous PCR0 committed in the EIF",
-		)
```
chris-ricketts: We still need to verify that `state.predecessorPCR0 == eifPreviousPCR0` - this enforces that the deployment lineage matches the reproducible build lineage.

runtime/config.go
```diff
@@ -5,12 +5,45 @@ import (
...
 	"os"
-	"strconv"
 	"strings"
+	"time"
 )
 
-// Config holds runtime HTTP/network settings.
+const (
+	prodRetention         = 10 * 365 * 24 * time.Hour
+	prodMigrationCooldown = 24 * time.Hour
```
chris-ricketts: This should not be hardcoded as it is application/deployment specific.

The intent of the migration cooldown is to allow users time to verify the proposed successor and to exit the application if they do not agree. 

A malicious host could turn off services that are required for a collaborative exit, therefore the cooldown must be long enough to allow for a unilateral exit of funds. This period is specific to the application and its deployment configuration.

runtime/runtime.go
```diff
@@ -164,8 +155,8 @@ func Run(ctx context.Context, cfg Config) error {
 	return supervise(ctx, rt, app)
 }
 
-func nsmFromEnv() NSM {
-	return NewNSM(WithAttestationUnsigned(skipCOSEVerification()))
+func nsmFromEnv(cfg *Config) NSM {
+	return NewNSM(WithAttestationUnsigned(cfg.InsecureVerifySkipped))
```
chris-ricketts: `cfg.InsecureVerifySkipped` is never set, always `false` breaks tests

runtime/config.go
```diff
@@ -75,33 +112,144 @@ func (c *Config) String() string {
...
-	if v == "" {
-		return fallback, nil
-	}
-	n, err := strconv.ParseUint(v, 10, 16)
-	if err != nil {
-		return 0, fmt.Errorf("%s=%q: %w", key, v, err)
+func (c *Config) setSecurityConfig(dev bool) {
+	c.Dev = dev
+	c.KMSLocked = !dev
+	c.VerifyClockSource = !dev
```
chris-ricketts: `VerifyClockSource` should be set by reading its own env var instead of tied to `ENCLAVE_DEV=true` - when testing with real AWS Nitro enclaves, we will want to verify the clock source but also use development S3 retention periods.

chris-ricketts (changes requested):
Posted a partial review. I noticed that the CI had never run and would fail due to `InsecureVerifySkipped` always being false.

CI is showing the failures @aruokhai - let me know when to continue the review.

runtime/boot.go
aruokhai: I don't think the deployment lineage needs to match the reproducible build lineage. We already have the migration transition receipt and the genesis artefact, which cover the same ground as far as I can tell. What does tying the two together actually buy us that those don't?.

aruokhai: Yes, after some thought, Our  authorization model is the  migration intent,. The successor should verify that authorization against the deployment log, alongside the transition receipt. We therefore don’t need to bind the predecessor into the successor’s build, but we do need to enforce intent verification on the successor.

chris-ricketts: The check ties state acceptance to the audited build chain: `B` must inherit from `A`, and `C` must inherit from `B`. A fully state-bound ancestry proof could replace it, but the current ancestry response relies on the inherited TLS key being trustworthy.

For example, we could have a genuine public genesis entry and intent log showing `A → B → C`, while the actual state came through `Z → B → C`, with `Z` supplying host-known secrets and a TLS private key. 

The host could relay the verifier’s fresh challenge to genuine `C`, return `C`’s valid attestation unchanged, then terminate TLS using that inherited key and return an honest-looking ancestry response. The nonce, PCR0, and TLS pin would all pass.

The predecessor check prevents that substitution at `B`. Removing it requires an equivalent binding from the live attestation to the actual state and its receipt chain back to the trusted genesis. 

Given the current design, I think retaining the EIF-committed predecessor check is cheaper and easier to reason about than implementing an equivalent state-bound ancestry proof.

chris-ricketts: > We therefore don’t need to bind the predecessor into the successor’s build, but we do need to enforce intent verification on the successor.

Can this be done in its own PR? The stated goal should be to remove the `state.predecessorPCR0 == eifPreviousPCR0` check with the justification that it simplifies the EIF build process by removing the required `ENCLAVE_PREVIOUS_PCR0` env var.

To cap the scope of this PR, it would be easier to simply restore the check.

runtime/runtime.go
```diff
@@ -82,15 +76,13 @@ func Run(ctx context.Context, cfg Config) error {
 
 	rt := newRuntimeState()
 
-	nsm := nsmFromEnv()
+	nsm := nsmFromEnv(&cfg)
```
chris-ricketts: Nit: inline helper function

```suggestion
	nsm := NewNSM(WithAttestationUnsigned(cfg.InsecureVerifySkipped))
```

chris-ricketts:
Added an additional E2E test script in b87984a38d25c6e5eff0506b003ce0dcd2453753

This regression test captures the consequence of removing the EIF-baked predecessor check: after a legitimate A → B migration, B accepts substituted state from an unauthorized predecessor Z under its existing KMS key.

B should reject Z before writing a new origin receipt or starting the application. Instead, it accepts the migration and becomes ready.

CI is intentionally red pending the predecessor-authorization fix.

cc @aruokhai

runtime/config.go
aruokhai: 43b1029d3fb084c17bf4e8beebeb03bba503b0ac

runtime/boot.go
aruokhai: 43b1029d3fb084c17bf4e8beebeb03bba503b0ac

runtime/config.go
aruokhai: c7ac9c8a0bb878c526e90e1592c83039f6fa1a6f

runtime/runtime.go
```diff
@@ -235,10 +225,6 @@ func supervise(ctx context.Context, rt RuntimeState, child appProcess) error {
 		_ = child.Stop()
 		return fmt.Errorf("HTTP listener failed: %w", err)
 
-	case <-rt.Halt():
-		slog.Info("received halt circuit-breaker")
-		return waitForRuntime(ctx, rt, child)
-
```
chris-ricketts: With the `Halt()` branch removed, the only caller to `waitForRuntime` is the `ChildDone()` branch. 

`waitForRuntime` can now be simplified to:

```go
// waitForRuntime keeps the runtime alive after the app has exited, so
// health and migration endpoints still answer.
func waitForRuntime(ctx context.Context, rt RuntimeState) error {
	select {
	case err := <-rt.ListenError():
		return fmt.Errorf("HTTP listener failed: %w", err)
	case <-ctx.Done():
		if cause := context.Cause(ctx); cause != nil && cause != context.Canceled {
			return fmt.Errorf("runtime halted: %w", cause)
		}
		slog.Info("shutting down")
		return nil
	}
}
```

runtime/logging.go
```diff
@@ -143,104 +37,19 @@ func generateLogID() string {
...
+// queryable history: CloudWatch is the record, and a truncated in-memory window
+// readable from the enclave would only be a weaker second one.
 type Logging struct {
-	buf     *logBuffer
-	shipCh  chan logEntry // nil when CloudWatch shipping is disabled
-	metrics *Metrics
-	cw      CloudWatchLogsAPI
+	telemetry *Telemetry
+	metrics   *Metrics
 }
```
chris-ricketts: Now that the `Logging` is just a wrapper around `*Telemetry` and `*Metrics` and is constructed only from `*Telemety`:

https://github.com/ArkLabsHQ/enclave/blob/b2538dc944daffe871eec6047962e0d5ad99b3e5/runtime/telemetry.go#L95

We can remove `Logging` and pass `*Telemetry` to `HandleLogsPost` and `NewSlogHandler`

runtime/servers.go
chris-ricketts: This field is never used

runtime/migration_intent.go
```diff
@@ -246,31 +266,79 @@ func (l *migrationIntentLog) deriveHead(
...
+	sequences := make([]uint64, 0, len(heads))
+	for sequence := range heads {
+		sequences = append(sequences, sequence)
 	}
-	return head, tie, nil
+	sort.Slice(sequences, func(i, j int) bool { return sequences[i] < sequences[j] })
+	var head *migrationIntent
+	var tie bool
+	var highest uint64
+	for _, sequence := range sequences {
```
chris-ricketts: We can use the stdlib `slices.Sorted` function.

```suggestion
	var head *migrationIntent
	var tie bool
	var highest uint64
	for _, sequence := range slices.Sorted(maps.Keys(heads)) {
```

runtime/migration_intent.go
```diff
@@ -228,11 +239,20 @@ func (l *migrationIntentLog) append(
 }
 
 func (l *migrationIntentLog) deriveHead(
+	ctx context.Context, sourcePCR0 string,
+) (*migrationIntent, bool, error) {
+	head, tie, _, err := l.scanIntents(ctx, sourcePCR0)
+	return head, tie, err
+}
```
chris-ricketts: We can remove this helper and change:

https://github.com/ArkLabsHQ/enclave/blob/b2538dc944daffe871eec6047962e0d5ad99b3e5/runtime/migration_intent.go#L110

to `head, tie, _, err := l.scanIntents(ctx, sourcePCR0)`

https://github.com/ArkLabsHQ/enclave/blob/b2538dc944daffe871eec6047962e0d5ad99b3e5/runtime/migration_intent.go#L223-L229

to `head, err := l.Head(ctx, sourcePCR0)`

runtime/boot.go
```diff
@@ -792,8 +925,22 @@ func writeStateReceipt(
 	ssm SSM,
 	stateRoot []byte,
 	purpose, param string,
+	opts ...SSMSetOption,
+) error {
+	return writeReceipt(ctx, nsm, ssm, stateOriginPayloadV1{
+		Purpose: purpose, StateRoot: stateRoot,
+	}, param, opts...)
```
chris-ricketts: We can delete this helper and just call `writeReceipt` directly in `WriteTransitionReceipt`

chris-ricketts (changes requested):

runtime/boot.go
aruokhai: af0de6497727a57b95565e57059b854f23e1a1a5

runtime/migration_intent.go
aruokhai: af0de6497727a57b95565e57059b854f23e1a1a5

runtime/migration_intent.go
aruokhai: af0de6497727a57b95565e57059b854f23e1a1a5

runtime/servers.go
aruokhai: af0de6497727a57b95565e57059b854f23e1a1a5

runtime/runtime.go
aruokhai: af0de6497727a57b95565e57059b854f23e1a1a5

runtime/runtime.go
aruokhai: af0de6497727a57b95565e57059b854f23e1a1a5

chris-ricketts (approved):
LGTM

PR #161, opened 2026-08-27
feat(runtime): automate attested blue-green migrations

aruokhai:
## Summary

This PR replaces the host-controlled migration API with an autonomous, attestation-based blue/green migration flow between enclaves.

Booting a successor candidate now acts as the migration request. The predecessor discovers it through SSM, verifies its live NSM attestation, records the migration intent, and commits the state handoff after the configured cooldown.

## What changed

- Boot successor enclaves in a restricted `candidate` state.
- Discover successors through an SSM challenge-response protocol.
- Derive the successor PCR0 directly from its verified NSM attestation.
- Promote candidates in place after the predecessor commits their state.
- Remove the unauthenticated migration HTTP endpoint and QEMU migration proxy.
- Add an operator-controlled SSM abort mechanism during the cooldown.
- Use an atomic create-only SSM write so horizontally scaled predecessors cannot overwrite an existing handoff.
- Ensure the expected predecessor PCR0 is measured into the successor EIF.
- Update documentation and E2E coverage for the new migration lifecycle.

## Migration flow

1. The operator builds and boots the successor EIF.
2. The successor starts as a candidate without application state or traffic.
3. The predecessor publishes a fresh challenge through SSM.
4. The candidate answers with an NSM attestation containing the challenge.
5. The predecessor verifies the attestation and records an Object-Locked migration intent.
6. After the cooldown, the predecessor re-encrypts the state for the successor PCR0 and atomically commits the handoff.
7. The successor verifies and adopts the state, then promotes itself without restarting.

If multiple candidates answer simultaneously, the predecessor treats the migration as ambiguous and does not proceed.

## Security model

Only the operator can deploy a candidate into the deployment environment, so booting that candidate is the authorization decision.

NSM attestation ensures the predecessor learns the genuine PCR0 of a live enclave rather than accepting a host-supplied value. The nonce prevents replay, while the atomic SSM commit ensures only one predecessor replica can finalize a handoff.

## Testing

- `make test`
- `go test -race ./...`
- `go vet ./...`
- `nix build .#checks.x86_64-linux.e2e --no-link --print-build-logs`

---
<a href="https://app.blacksmith.sh/ArkLabsHQ/codesmith/enclave/pr/161"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790409875&installation_model_id=433183&pr_number=161&repository=ArkLabsHQ%2Fenclave&return_to=https%3A%2F%2Fgithub.com%2FArkLabsHQ%2Fenclave%2Fpull%2F161&signature=edd9c811b18cda9bfd4f29c74c7492fe4818b8bf5290c27c0003bd1042dd78fd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/runtime/migrate.go
+++ b/runtime/migrate.go
@@ -1,22 +1,41 @@
 package runtime
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"log/slog"
 	"math"
+	"strings"
 	"sync"
 	"time"
+
+	"github.com/fxamacker/cbor/v2"
 )
 
 // migrationPCRIndex stores the successor-PCR0 handoff commitment.
-const migrationPCRIndex = 31
+const (
+	migrationPCRIndex        = 31
+	migrationPollInterval    = 5 * time.Second
+	migrationChallengeRotate = time.Minute
+	successorClaimSchemaV1   = "enclave.successor_claim.v1"
+)
 
-type MigrationRequest struct {
-	Action     string `json:"action"`
-	TargetPCR0 string `json:"target_pcr0,omitempty"`
+// successorClaimV1 is the attested pre-image a candidate offers. Deployment is
+// measured and not SSM-overridable, so it binds the claim to one deployment and
+// a document produced for another is rejected.
+type successorClaimV1 struct {
+	Schema     string `cbor:"schema"`
+	Deployment string `cbor:"deployment"`
+}
+
+// CandidateInfo is reported by an enclave still awaiting a handoff.
+type CandidateInfo struct {
+	AwaitingHandoffFrom string     `json:"awaiting_handoff_from,omitempty"`
+	RequestedAt         *time.Time `json:"requested_at,omitempty"`
 }
 
 type CompleteMigrationResult struct {
@@ -48,13 +67,17 @@ const (
 )
 
 type Migrator interface {
-	HandleMigrationRequest(
-		ctx context.Context,
-		action, targetPCR0 string,
-	) (*MigrationStatus, error)
-	CompleteMigration(ctx context.Context) (*CompleteMigrationResult, error)
+	// RunMigrationControl drives the predecessor half, for the life of a
+	// serving enclave.
+	RunMigrationControl(ctx context.Context)
+	// RunCandidateAttestation drives the candidate half, until promotion.
+	RunCandidateAttestation(ctx context.Context)
+
 	PreviousPCR0Info(ctx context.Context) (*PreviousPCR0Info, error)
 	MigrationStatus(ctx context.Context) (*MigrationStatus, error)
+	CandidateInfo(ctx context.Context) (*CandidateInfo, error)
+	Ready() bool
+	Promote(kms PrimaryKMS, dek DEK, secrets []StaticSecret)
 }
 
 type migrator struct {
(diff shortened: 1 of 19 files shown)


runtime/migrate.go
```diff
@@ -115,108 +153,139 @@ func (m *migrator) PreviousPCR0Info(ctx context.Context) (*PreviousPCR0Info, err
...
+//	  commit KMSKeyID/<target>  ------->  promote
+//
+
+func (m *migrator) RunPredecessorHandoff(
+	ctx context.Context, kms PrimaryKMS, dek DEK, secrets []StaticSecret, tlsKey crypto.Signer,
+) {
+	m.mu.Lock()
+	m.kms, m.dek, m.staticSecrets, m.tlsKey = kms, dek, secrets, tlsKey
+	m.mu.Unlock()
+	m.promoted.Store(true)
```
chris-ricketts: We don't need this synchronisation, i.e. `mu` and `promoted` 

The fields assigned under the mutex are only read within this goroutine, promoted is only read in `advanceMigration` and never set back to false.

runtime/migrate.go
```diff
@@ -51,62 +68,83 @@ const (
...
 	kms           PrimaryKMS
 	ssm           SSM
 	dek           DEK
 	staticSecrets []StaticSecret
 	tlsKey        crypto.Signer
 	intent        *migrationIntentLog
+	genesis       *genesisLog
+	promoted      atomic.Bool
+
+	challenge         atomic.Pointer[migrationChallenge]
```
chris-ricketts: I don't think this needs to be an atomic pointer, it's never read/set concurrently

runtime/migrate.go
```diff
@@ -246,117 +315,316 @@ func (m *migrator) verifyIntent(
...
+	if err != nil {
+		return nil, fmt.Errorf("read migration challenge: %w", err)
+	}
+	var challenge migrationChallenge
+	if json.Unmarshal([]byte(published), &challenge) == nil {
+		age := time.Since(challenge.IssuedAt)
+		if challenge.Nonce != "" && age >= 0 && age < migrationChallengeRotate {
+			return &challenge, nil
+		}
+	}
```
chris-ricketts: An expired challenge response could be replayed by an entity with SSM write permissions that writes and expired nonce with a fresh issued at timestamp.

We could replace the challenge nonce and timestamp with an attestation document that contains the nonce and issued timestamp.

runtime/migrate.go
```diff
@@ -246,117 +315,316 @@ func (m *migrator) verifyIntent(
...
+	if target == "" {
+		return nil
+	}
+
+	if err := lease.Verify(ctx); err != nil {
+		return fmt.Errorf("verify migration lease: %w", err)
+	}
+	if _, err := m.intent.Request(ctx, m.pcr0, target); err != nil {
+		return err
+	}
```
chris-ricketts: After accepting a response and recording an intent, we should retire the consumed challenge and issue a fresh one when looking for a successor again.

If the operator stops and aborts the candidate before the challenge’s one-minute expiry, the predecessor resumes looking for a successor. The aborted candidate’s stored response still answers the current challenge, so the predecessor records another migration request without any new response from that candidate.

chris-ricketts (changes requested):

runtime/migrate.go
aruokhai: 530a9707b4144106c6a6db830873c4d43a963ca1

runtime/migrate.go
aruokhai: 530a9707b4144106c6a6db830873c4d43a963ca1

runtime/migrate.go
aruokhai: 530a9707b4144106c6a6db830873c4d43a963ca1

chris-ricketts (approved):
LGTM

PR #165, opened 2026-08-27
fix(runtime): stop claiming the whole /v1 subtree on the external mux

### LINKED ISSUE #164: Runtime claims the entire /v1/ subtree, making an application's own /v1/ routes unreachable
The runtime claims the entire `/v1/` subtree, so an application's own `/v1/*` routes are
unreachable — they 404 at the runtime instead of reaching the reverse proxy.

[`runtime/servers.go:106-108`](https://github.com/ArkLabsHQ/enclave/blob/d42c1d7/runtime/servers.go#L106-L108):

```go
em.Handle("/v1/",    corsWildcard(sm))   // subtree pattern: matches all of /v1/*
em.Handle("/health", sm)
em.Handle("/",       revProxy)           // the application
```

`sm` serves only `/v1/metrics`, `/v1/logs`, `/v1/traces` and the `/v1/enclave-*` reads.
Anything else under `/v1/` returns 404 rather than falling through to `revProxy`.

**Observed.** The emulator's whole REST surface is under `/v1/`, and none of it is
reachable through the enclave:

```
GET  /v1/info           404
POST /v1/tx             404
POST /v1/intent         404
POST /v1/finalization   404
POST /v1/onchain-tx     404
```

Its gRPC works, because the path is `/emulator.v1.EmulatorService/GetInfo`. `/healthz`
works, because it is outside `/v1/`. Only the REST gateway is dead.

`/v1/` is a conventional API prefix, so this likely affects any application with a
versioned REST API, not just this one. It is also not configurable around — the prefix is
hardcoded and no environment variable affects routing.

**Options:**

1. Register the runtime's `/v1/` endpoints individually. They are already enumerated at
   [`servers.go:94-100`](https://github.com/ArkLabsHQ/enclave/blob/d42c1d7/runtime/servers.go#L94-L100),
   so the subtree pattern buys nothing.
2. Move the runtime's API under `/enclave/v1/`, matching the `/enclave/*` endpoints it
   already owns.
3. Fall through to `revProxy` when `sm` would 404.

Option 1 is the smallest change and preserves every existing path.

s373nZ:
Fixes #164.

`em.Handle("/v1/", corsWildcard(sm))` is a Go `ServeMux` **subtree** pattern, so every
`/v1/*` request was routed to `sm` whether or not `sm` knew the path. An application
serving its own versioned API under `/v1` had all of it answered by the runtime's 404,
and never reached `revProxy`.

## Observed

Against a real deployment, the emulator's entire REST surface was unreachable through the
enclave:

```
GET  /v1/info           404
POST /v1/tx             404
POST /v1/intent         404
POST /v1/finalization   404
POST /v1/onchain-tx     404
```

Its gRPC worked, because that path is `/emulator.v1.EmulatorService/...`, and `/healthz`
worked because it sits outside `/v1`. Only the REST gateway was dead — which is what made
it easy to miss.

`/v1` is a conventional API prefix, so this affects any application with a versioned REST
API, and it is not configurable around: the prefix is hardcoded and no environment
variable touches routing.

## Change

Register the runtime's six `/v1` endpoints individually. They were already enumerated on
`sm`, so the subtree pattern was not buying anything.

Two details preserved deliberately:

- **`/v1/enclave-info` preflight.** It is registered later by
  `ConfigureEnclaveInfoHandler` as `GET`-only, and relied on the subtree to answer its
  `OPTIONS`. That route is now explicit, so CORS behaviour is unchanged.
- **The internal mux keeps its subtree.** `im` has no reverse proxy, so there is nothing
  for it to shadow.

Paths are registered without a method, as the subtree was, so `OPTIONS` still reaches
`corsWildcard` and a method mismatch still yields 405 from `sm` rather than falling
through to the application.

## Test

`TestExternalMuxLeavesApplicationV1RoutesToTheProxy` stands up a real `SetupHttpServers`
against a recording backend and asserts three things: application routes under `/v1`
reach the proxy, the runtime's own routes do not, and preflight is still answered for the
runtime's routes including `/v1/enclave-info`.

Confirmed it fails on the old routing with exactly the reported symptom:

```
--- FAIL: TestExternalMuxLeavesApplicationV1RoutesToTheProxy/application_routes_reach_the_proxy
        expected: 200
        actual  : 404
        Messages: /v1/info
```

`go test ./...`, `go vet ./...` and `gofmt -l .` are clean on the full runtime module.

---
<a href="https://app.blacksmith.sh/ArkLabsHQ/codesmith/enclave/pr/165"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790418593&installation_model_id=433183&pr_number=165&repository=ArkLabsHQ%2Fenclave&return_to=https%3A%2F%2Fgithub.com%2FArkLabsHQ%2Fenclave%2Fpull%2F165&signature=0ec7f6f409ac97fd73adbf8d30fddcde7de433a798b89228a1a26232596b56f8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,6 @@ measurement. A subset can be overridden at runtime from SSM.
 | `ENCLAVE_NITRIDING_UPSTREAM` | `auto` | Runtime-to-application HTTP version. `h1` pins HTTP/1.1, `h2c` pins HTTP/2 cleartext and is required for gRPC, `auto` matches the inbound request. |
 | `ENCLAVE_NITRIDING_FQDN` | `localhost` | Hostname for the TLS certificate. |
 | `ENCLAVE_NITRIDING_HOST_PROXY_PORT` | `1024` | Host vsock port where gvproxy listens. |
-| `ENCLAVE_NITRIDING_DEBUG` | `false` | Reported by `GET /enclave/config`. |
 | `ENCLAVE_VIPROXY_ENABLED` | `true` | Set to `false` to disable the in-process IMDS forwarder. |
 | `ENCLAVE_VIPROXY_IN_ADDRS` | `127.0.0.1:80` | IMDS forwarder listen address. |
 | `ENCLAVE_VIPROXY_OUT_ADDRS` | `3:8002` | IMDS forwarder target, `CID:PORT` or `host:port`. |
@@ -381,13 +380,13 @@ With `D` = deployment, `A` = app name, `L` = `locked` or `unlocked`:
 
 TLS 1.2 minimum. Every non-gRPC response carries `X-Attestation-Signature` and
 `X-Attestation-Pubkey`, a BIP-340 Schnorr signature over the SHA-256 of the
-response body. `/enclave/v1/*` responses also carry permissive CORS headers.
+response body. `/enclave/*` responses also carry permissive CORS headers, and
+an `OPTIONS` preflight to any path in that namespace is answered `204` by the
+runtime.
 
 | Method | Path | Auth | Purpose |
 |---|---|---|---|
 | GET | `/enclave/attestation?nonce=<40 hex>` | none | NSM attestation document, base64. The nonce is mandatory and echoed back. |
-| GET | `/enclave` | none | Human-readable index. |
-| GET | `/enclave/config` | none | Effective runtime configuration as JSON. |
 | GET | `/enclave/v1/info` | none | Version, PCR0, predecessor PCR0 and attestation, attestation public key, migration status, application status. |
 | GET | `/health` | none | `{"status":"ready"}` once the application has been started, `{"status":"initializing"}` with status 503 before. |
 | GET | `/enclave/v1/metrics` | none | Metric snapshot. |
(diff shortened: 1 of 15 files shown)


s373nZ:
I'm still a little curious about Option 2 from #164:
> 2. Move the runtime's API under /enclave/v1/, matching the /enclave/* endpoints it
already owns.

or, running the enclave service on a separate port if possible, but this seems like an easy fix.

aruokhai:
> I'm still a little curious about Option 2 from #164:
> 
> > 2. Move the runtime's API under /enclave/v1/, matching the /enclave/* endpoints it
> >    already owns.
> 
> or, running the enclave service on a separate port if possible, but this seems like an easy fix.

I think i prefer the second option, much cleaner

aruokhai:
@s373nZ what do you think about the changes

s373nZ (commented):
@aruokhai Looks pretty good to me. I got confused there for a minute becuase the tests were using `/v1/logs` and `/v1/metrics` as application routes alongside `/enclave/v1/logs` etc, but I get it now.

